Issue #4, opened 2026-04-14
fails to compile on arch linux


donington:
Based off the error I receive:
```
error: library kind `framework` is only supported on Apple targets
error: could not compile `accelerate-src` (lib) due to 1 previous error
```
I assume that you must be developing on mac hardware, since from what I can find `accelerate-src` is dependent on apple/metal framework.  If you want this to meet a broader audience I think there's merit in stating the current limitations up front to get better reception.  Also, very cool concept and I wish I could launch it to tinker with the actual approach.  I will explore further but this is a language I am not familiar with so I cannot provide greater feedback than it didn't work on my platform.

codeninja:
https://github.com/chrishayuk/larql/pull/6 might fix you?

donington:
This did work!  The only additional thing I had to set up was a venv to target python 3.13 specifically in order for pyo3-ffi to compile.

codeninja:
I didn't run into the 3.13 issue personally. But I did Identify a few additional fixes in PR #8 which you might run into.

donington:
Everything is now working as expected, closing this.

PR #5, opened 2026-04-14
fixed compilation not working by adding selection for correct backend

SpezialK-dev:
This should fix the error mentioned in #4 also specified that OpenBlas is needed under Linux in the Readme file for Linux. 

I could only test this under Linux and it compiled fine so Tests under Mac are still needed.


codeninja:
See PR https://github.com/chrishayuk/larql/pull/8 for additional fixes.

PR #6, opened 2026-04-14
fix: Linux/WSL2 support (OpenBLAS, GGUF vocab_size + transpose, lql multi-statement) + --temperature parameter

kariboo84:
## Summary

Four bug fixes to run LARQL on Linux/WSL2 + one new feature.
Tested on Windows 11 / WSL2 Ubuntu, i5-12600K, 32GB RAM, google/gemma-3-4b-it safetensors.

## Bug fixes

**1. Replace `accelerate-src` with OpenBLAS**
`accelerate-src` is Apple-only and fails to compile on Linux.
Replaced with `openblas-src` (`system` feature) in `larql-compute` and `larql-inference`.

**2. Fix GGUF `vocab_size=0` on Linux**
Gemma 3 GGUF returns `vocab_size=0` from arch metadata on Linux.
Now reads vocab_size from `tokenizer.json` as fallback (262144).

**3. Fix GGUF embedding transpose on Linux**
GGUF loads embeddings as `[hidden_size, vocab_size]` instead of `[vocab_size, hidden_size]`.
Auto-transposes when `shape[0] < shape[1]`.

**4. Fix `lql` multi-statement execution**
`larql lql 'USE "..."; DESCRIBE "France";'` silently dropped everything after the first statement.
One-line fix: replace `run_statement` with `run_batch` in the CLI dispatcher.

## New feature

**5. Add `--temperature` to `predict`**
- Default `1.0` — no change to existing behavior
- `T=0.1` → Paris 100% (sharp)
- `T=2.0` → Paris 9% (flat)
- Applied in `logits_to_predictions` before softmax

### DIFF
--- a/crates/larql-cli/src/commands/extraction/predict_cmd.rs
+++ b/crates/larql-cli/src/commands/extraction/predict_cmd.rs
@@ -1,5 +1,6 @@
 use std::time::Instant;
 
+use larql_inference::forward::predict_with_temperature;
 use clap::Args;
 use larql_inference::{
     calibrate_scalar_gains, predict, predict_with_ffn, predict_with_router, predict_with_strategy,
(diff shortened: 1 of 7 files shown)


codeninja:
I'm also running this setup, I'll test this branch when I can.

codeninja:
Should the Temperature=1.0 preserve pre-PR behavior (no softmax scaling) in the  Predict.rs & runner.rs call to logits_to_predictions_pub?

codeninja:
See PR #8 for additional fixes.

PR #24, opened 2026-04-17
feat(gemma4): Add Gemma 4 GGUF support + fix column-major loading and Q4_K dequantization

mikeumus:
## Summary

Adds support for Google's Gemma 4 models via GGUF. **Also fixes three pre-existing correctness bugs** that silently affected every GGUF model LarQL has loaded.

**First-ever successful LarQL extraction of Gemma 4** (e2b, 5.1B params, 35 layers, per-layer variable FFN: 6144→12288).

### 1. GGUF column-major loading (critical — affects ALL GGUF models)

GGUF/GGML stores 2D tensors in **column-major** (Fortran) layout where `dims[0]` = columns and `dims[1]` = rows. The loader was interpreting data as row-major, producing **transposed matrices** for every GGUF model.

Silently wrong for most models; only exposed as a hard panic when Gemma 4's variable per-layer FFN caused the transposed matmul to fail:

```
ndarray: inputs 262144 × 1536 and 6144 × 1024 are not compatible
```

**Fix**: Create ndarray with Fortran layout (`.f()`) using swapped dims `(ne1, ne0)`, then convert via `.as_standard_layout()`.

### 2. Q4_K dequantization (two bugs, affects Q4_K_M models including Gemma 4)

- **Wrong block size**: 148 bytes declared; actual Q4_K block is **144 bytes** (2 + 2 + 12 + 128 per llama.cpp's `block_q4_K`).
- **Wrong scale unpacking**: read bytes 4-11 as 8 individual 6-bit scales, but Q4_K packs 8 scales + 8 mins into 12 bytes (4-15) via llama.cpp's `get_scale_min_k4` scheme.

Result: Q4_K-quantized weights produced garbage → NaN in attention SVD (`qk-rank`, `ov-gate`).

**Fix**: Rewrote per llama.cpp reference. For `j<4`: extract 6 bits from `scales[j]` and `scales[j+4]`. For `j≥4`: combine lower 4 bits of `scales[j+4]` with upper 2 bits from `scales[j-4]` and `scales[j]`.

### 3. Gemma 4 architecture support

- Added `gemma4` to GGUF→HF architecture mapping
- `get_arch_u32` now handles array metadata (takes max) for Gemma 4's per-layer variable FFN
- Override `head_dim=256` for Gemma 4 (its `attention.key_length=512` reports a different dim than per-head)

### 4. NaN-safe sorting

Replaced `partial_cmp().unwrap()` with `unwrap_or(Ordering::Equal)` in `build.rs` and `qk_rank_cmd.rs`. Prevents panics when quantized weights produce NaN.

## Testing

Gemma 4 (e2b, 5.1B dense) extracts successfully:
- 35 layers, per-layer variable FFN sizes preserved in `index.json`
- Vindex size: 1.8GB (gate_vectors 1.0GB, embeddings 768MB at f16)
- Gate bytes match math exactly: `(15×6144 + 20×12288) × 1536 × 2 = 1,038,090,240`

After Q4_K fix, `qk-rank` on Gemma 4 produces real numerical singular values (was returning NaN for all 24 heads before).

## Validation beyond compilation

Used on a research project mapping dense FFN sparsity to MoE expert count predictions. Q4_K fix was validated by comparing Gemma 4 attention SVD output (previously all NaN, now meaningful singular values with a consistent pattern across layers).

## Notes

- The column-major fix applies to every GGUF model already. Qwen / Llama / Mistral GGUFs were silently loading transposed weights; their extraction "worked" because the downstream math still got compatible shapes — but the stored gate vectors were not what the model's forward pass actually uses.
- Gemma 4 vision/audio blocks (`v.blk.*`, `a.blk.*`) are naturally filtered out by the existing tensor key normalizer — they don't match the text block pattern.

### DIFF
--- a/crates/larql-cli/src/commands/extraction/qk_rank_cmd.rs
+++ b/crates/larql-cli/src/commands/extraction/qk_rank_cmd.rs
@@ -252,7 +252,7 @@ fn compute_singular_values(ata: &ndarray::Array2<f32>, dim: usize) -> Vec<f32> {
         matrix = matrix - outer;
     }
 
-    eigenvalues.sort_by(|a, b| b.partial_cmp(a).unwrap());
+    eigenvalues.sort_by(|a, b| b.partial_cmp(a).unwrap_or(std::cmp::Ordering::Equal));
     eigenvalues
 }
 
--- a/crates/larql-models/src/loading/gguf.rs
+++ b/crates/larql-models/src/loading/gguf.rs
@@ -8,7 +8,7 @@ use std::collections::HashMap;
 use std::io::{BufReader, Read, Seek};
 use std::path::Path;
 
-use ndarray::Array2;
+use ndarray::{Array2, ShapeBuilder};
 
 use crate::weights::ModelWeights;
 use crate::detect::ModelError;
@@ -194,11 +194,23 @@ impl GgufFile {
 
             match info.n_dims {
                 2 => {
-                    // GGUF stores in row-major, dims[0] = rows, dims[1] = cols
-                    let rows = info.dims[0] as usize;
-                    let cols = info.dims[1] as usize;
-                    let arr = Array2::from_shape_vec((rows, cols), floats)
+                    // GGUF/GGML uses column-major (Fortran) dimension ordering:
+                    //   dims[0] = number of columns (innermost/fastest)
+                    //   dims[1] = number of rows (outermost)
+                    // Data is laid out in column-major order.
+                    //
+                    // ndarray expects row-major (C) order by default.
+                    // To get the correct [rows, cols] matrix in row-major ndarray,
+                    // we swap the dimensions and use Fortran (column-major) layout,
+                    // then convert to standard (C) layout via .as_standard_layout().
+                    let ne0 = info.dims[0] as usize; // columns in GGML
+                    let ne1 = info.dims[1] as usize; // rows in GGML
+                    // Shape is (rows, cols) = (ne1, ne0) in standard math convention.
+                    // Data is column-major, so we create with Fortran layout.
+                    let arr = Array2::from_shape_vec((ne1, ne0).f(), floats)
                         .map_err(|e| ModelError::Parse(format!("tensor {}: {}", info.name, e)))?;
+                    // Convert to standard (C/row-major) layout for compatibility
+                    let arr = arr.as_standard_layout().into_owned();
                     tensors.insert(key, arr.into_shared());
                 }
                 1 => {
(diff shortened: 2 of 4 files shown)


mikeumus:
## Update — fixes validated end-to-end

Since opening this PR, I've further validated these fixes:

**1. Vindex published** for community use: https://huggingface.co/datasets/Divinci-AI/gemma-4-4b-e2b-vindex (1.8 GB)

**2. Numerical validation of Q4_K fix**: Ran `qk-rank` on Gemma 4 with the Q4_K patch. Before the fix, all 24 heads returned `rank=0, s_max=NaN`. After the fix, real numerical singular values across all 35 layers (s_max range 1.8-3.6, consistent U-shape-alternative profile).

**3. Per-layer variable FFN verified correct**: L0-14 have 6144 features, L15-34 have 12288. Gate vectors file is exactly `(15×6144 + 20×12288) × 1536 × 2 bytes = 1,038,090,240` bytes, matching the expected byte count for correctly-loaded dimensions.

**4. Cross-scale research application**: Used this to validate a scaling law across Qwen 0.5B, Qwen 1.5B, SmolLM2 360M, and Gemma 4. Dense FFN dominant-feature sparsity predicts MoE expert count within 4% (Gemma 4 measured: 12% sparsity → predicts 8 experts, actual: top-8 of 128).

**5. Also confirmed downstream via Cloudflare Workers AI logprobs**: top-8 tokens capture 99.7% of Gemma 4 26B MoE output probability mass — the "8" in the architecture shows up at both the structural (LarQL) and behavioral layers.

The column-major fix is the critical item for any non-Gemma-4 user — it's a silent correctness bug affecting every GGUF model LarQL has loaded. Worth reviewing even if the Gemma 4 additions aren't a priority for you.

Happy to split into smaller PRs if that's easier to review.

chrishayuk:
sweeet. this is super useful....
i've just got gemma 4 working on one of my branches already...
i will merge with what i've got and see how it converges
thank you so much

chrishayuk:
@mikeumus really nice. thank you, merged now. super helpful

mikeumus:
@chrishayuk no thank you for publishing this. The magnitude of this is huge. We are grateful 🙇🏻‍♂️💕

We made a vindex of Gemma4 here if you want to try it out:
https://huggingface.co/Divinci-AI/gemma-4-4b-e2b-vindex

Also our fork of your LarQL has some fun new things in it here you're welcome to pull back into `upstream/main` here or I can create more PRs!
https://github.com/Divinci-AI/larql

I'm probing Gemma4 deeply in many ways with LarQL and journaling every step of the experiment. Will publish a blog post on it once we have something to show.

PR #35, opened 2026-04-26
feat(safetensors): support F8_E4M3 / F8_E5M2 / F8_E8M0 / I8 dtypes

mikeumus:
## Summary

Bumps `safetensors` 0.5 → 0.7 and adds bit-pattern → f32 decoders for four dtypes that the current loader rejects with `parse error: InvalidHeaderDeserialization`:

- **`F8_E8M0`** — Open Compute MX-format scale (per-32-element, value `2^(byte − 127)`). Required by every MXFP4-quantized model.
- **`F8_E4M3` / `F8_E5M2`** — standard FP8 weights (Open Compute Project FN encoding). Used by GPT-OSS, DeepSeek-V3 attention weights, etc.
- **`I8`** — signed bytes; in the MXFP4 case these are packed FP4 nibbles (the `.weight` companion of an `F8_E8M0` `.scale`).

This unblocks loading any safetensors file from the **DeepSeek-V4 family** (`deepseek-ai/DeepSeek-V4-Flash`, `…-Pro`, and unsloth repacks of either) and similar MXFP4-quantized MoE models.

## What's NOT in this PR (intentional)

- **MXFP4 cross-tensor unpacking.** `tensor_to_f32` here decodes each tensor in isolation. Proper MXFP4 dequantization (pairing an `I8` packed-nibble weight with its `F8_E8M0` scale companion) belongs at the FFN tensor loading layer where both tensors are loaded together — not in the per-tensor dispatch. A code comment marks this as the next layer to extend (pointing at `crates/larql-models/src/quant/mxfp4.rs` which already has the unpack primitive).

## Drive-by build fixes

While here, two small unrelated fixes so `cargo build --no-default-features` is clean on Linux from a fresh clone:

1. **`bench_cmd.rs`**: strip the `metal`-only `MetalBackend::new()` branch so the file compiles when the `metal` feature is off (the `metal` crate is `cfg(target_os = \"macos\")`-only). Bench is unaffected — it always used `CpuBackend` outside of macOS anyway.
2. **`compile_cmd/save.rs`**: update `safetensors::serialize(&views, &None)` → `serialize(&views, None)` — safetensors 0.7 took the second argument by value instead of by reference.

## Verification

\`\`\`
cargo check -p larql-cli --no-default-features --bin larql
# → Finished `dev` profile [unoptimized + debuginfo] target(s)
# (1 unrelated dead-code warning on hf_api_url)
\`\`\`

Tested against deepseek-ai/DeepSeek-V4-Flash safetensors snapshot — `larql extract` now gets past the header parse and into the actual weight loading.


chrishayuk:
Forward-ported by #74 to land on top of current main. Authorship on the rebased commit preserved as you (`Mike Mooring <mike@divinci.ai>`).

The original branch carried ~17k LOC of Divinci-AI fork drift (your RFC-0001 commands `larql crown`/`edit`/`memit`, Python bindings, fork-specific docs) on top of an April-24 base. Only the `01e5f0d` feature commit was carried over to #74; the .bak.bak sed leftovers (which your own follow-up commit cleaned up) are gone, and unrelated drift in `bench_cmd.rs` was kept at main's version.

Thanks for the FP8/I8 dtype work — it's the foundation for #37/#39/#40, which we'll forward-port next.

PR #36, opened 2026-04-26
feat(hf): try Dataset → fall back to Model repo type when fetching vindex

mikeumus:
## Summary

vindexes can live in HF **model** repos as well as dataset repos, but the resolver hardcodes `RepoType::Dataset` everywhere. Pulling from a model-type vindex repo currently fails with:

\`\`\`
Error: parse error: failed to fetch index.json from hf://owner/name
\`\`\`

…even though `index.json` is right there in the repo. This PR adds a Dataset → Model fallback to the three resolve/download functions so both layouts work.

## What changed

In `crates/larql-vindex/src/format/huggingface.rs`, patches:

- `resolve_hf_vindex` — non-progress entrypoint
- `resolve_hf_vindex_with_progress` — the path the CLI's `pull` uses
- `download_hf_weights` — lazy weight-file fetch

Each function now:
1. Builds the repo handle via a `make_repo` closure that takes either `RepoType::Dataset` or `RepoType::Model`.
2. Tries Dataset first (preserves legacy default — most existing vindex repos remain dataset-type).
3. On any error fetching `index.json`, retries with Model.
4. All subsequent file fetches in the same call use the type that succeeded.

If both fail, the error message lists both attempts:

\`\`\`
failed to fetch index.json from hf://Divinci-AI/kimi-k2-instruct-vindex:
  tried as dataset (404 Not Found); tried as model (404 Not Found)
\`\`\`

## Verification

\`\`\`bash
$ larql pull Divinci-AI/kimi-k2-instruct-vindex
Pulling hf://Divinci-AI/kimi-k2-instruct-vindex...
Cached at: ~/.cache/huggingface/hub/models--Divinci-AI--kimi-k2-instruct-vindex/snapshots/7ca3228...
\`\`\`

The `models--` cache prefix confirms the fallback correctly selected Model. Before this PR the same command 404'd at the dataset URL.

## Why this matters

Three real-world vindex repos that need this path:
- `Divinci-AI/kimi-k2-instruct-vindex` (Kimi-K2-Instruct, fp8 MoE)
- `Divinci-AI/deepseek-v4-flash-vindex` (DeepSeek-V4-Flash, MXFP4 MoE)
- `Divinci-AI/deepseek-v4-pro-vindex` (DeepSeek-V4-Pro, MXFP4 MoE)

These were published under `repo_type="model"` (the upstream HF default for weight-bearing repos). Without this PR, none of them are `larql pull`-able.

## Companion PR

#35 (F8_E4M3 / F8_E5M2 / F8_E8M0 / I8 dtypes) unblocks `larql extract` on these same models. With both PRs merged, the full produce-and-consume flow for MXFP4 MoE vindexes works end-to-end.


chrishayuk:
Superseded by #71 (merged). Both PRs add HF repo-type fallback for vindex pulls, but #71 uses the order that matches `larql publish`'s default — Model first, then Dataset — whereas this PR's order is Dataset → Model. Since publish defaults to model repos (`publish/mod.rs:47`), the model-first order is what users actually need. Thanks for spotting the bug independently — your validation against a model repo helped confirm the fix was needed.

The other 7 PRs in this stack (#35, #37, #38, #39, #40, #42, #43) are still under review — we're working through them one at a time, cherry-picking each feature commit onto current main to drop the stale-fork noise and preserve your authorship.

PR #37, opened 2026-04-26
feat(mxfp4): per-expert dequantization for DeepSeek-V4 layout

mikeumus:
## Summary

Adds MXFP4 dequantization for the **per-expert, per-projection** layout used by DeepSeek-V4 (and likely other newer MoE models). DeepSeek-V4 stores expert weights as one `(weight, scale)` pair per `(expert, projection)`:

\`\`\`
layers.X.ffn.experts.E.w1.weight  dtype=I8        shape=[2048, 128]   ← packed FP4 nibbles
layers.X.ffn.experts.E.w1.scale   dtype=F8_E8M0   shape=[2048, 64]    ← per-32-element scales
\`\`\`

This is distinct from the **fused** `experts.gate_up_proj_blocks` + `experts.gate_up_proj_scales` layout used by GPT-OSS, which the existing `dequantize_mxfp4_experts` function handles.

## What changed

`crates/larql-models/src/loading/safetensors.rs`: new function `dequantize_per_expert_mxfp4` that:

1. Pattern-matches on tensor names ending in `.experts.<digit>.w[123].weight` with `I8` dtype.
2. Looks up the companion `.scale` tensor (`F8_E8M0`).
3. Validates shape consistency (`weight.cols * 2 == scale.cols * 32`).
4. Dequantizes via the existing `quant::mxfp4::dequantize_expert` primitive (no new core math).
5. Returns the set of consumed tensor names so the main loop skips them.

Wired into the non-PackedMxfp4 branch of `load_model_dir_filtered` so it runs alongside the regular weight-loading path. No-op when no V4-style expert tensors are present (cost: one set lookup + one dtype check per tensor).

## Verified

\`\`\`
cargo check -p larql-cli --no-default-features
# Finished `dev` profile [unoptimized + debuginfo] target(s)
\`\`\`

End-to-end test against `unsloth/DeepSeek-V4-Flash` is the next step — running `larql extract` should now load the expert weights instead of skipping them.

## Stacks on

- #35 (F8_E4M3/E5M2/E8M0/I8 dtype dispatch) — required for the safetensors header to parse at all
- #36 (Dataset → Model fallback) — required for `larql pull` of model-type vindex repos


chrishayuk:
Forward-ported by #75 to land on top of current main + the FP8/I8 dtype support from #74 (your #35 → #74). Authorship on the rebased commit preserved as you (`Mike Mooring <mike@divinci.ai>`).

Conflict notes carried over to #75's body — the main change vs. your original is integrating the V4 detector with main's PackedBF16 (Gemma 4 26B A4B) raw-bytes path so both coexist, and adding `?` propagation since `quant::mxfp4::dequantize_expert`'s signature on current main returns `Result`. The detector logic itself is untouched from your commit.

PR #38, opened 2026-04-26
feat(hf): metadata-only resolve_hf_vindex (no eager binary downloads)

mikeumus:
## Summary

\`resolve_hf_vindex\` previously fetched the entire \`VINDEX_CORE_FILES\` list eagerly, including \`gate_vectors.bin\` and \`embeddings.bin\` — multi-GB binaries that metadata-only commands like \`larql show\` never read. On a 42 GB Kimi-K2 vindex, \`larql show\` would hang for 10+ minutes pulling tensors it doesn't display.

## Change

Splits \`VINDEX_CORE_FILES\` into two lists:

- **\`VINDEX_METADATA_FILES\`** — small json/manifests/tokenizer/down_meta. Pulled eagerly by \`resolve_hf_vindex\`. Sub-second on cached repos.
- **\`VINDEX_BIN_FILES\`** — \`gate_vectors.bin\` + \`embeddings.bin\`. Deferred to call sites that actually need them (\`run\`, \`walk\`, etc.).

The progress-aware variant (\`resolve_hf_vindex_with_progress\`) keeps the prior eager behavior via a \`vindex_core_files()\` helper that returns the union — its caller has explicitly opted into a progress bar, so they accept the wait. File coverage there is byte-for-byte identical.

## Verification

\`\`\`
$ larql show Divinci-AI/kimi-k2-instruct-vindex   # 42 GB vindex, fresh cache
Model:      Divinci-AI/kimi-k2-instruct-vindex
Layers:     61
Hidden:     7168
Dtype:      F32
Quant:      None
\`\`\`

Returns in ~1s now instead of hanging on the bin download.

## Stacks on

#36 (Dataset → Model fallback). PRs #35 and #37 are independent of this one.


chrishayuk:
Forward-ported by #79. Authorship on the rebased commit preserved as you (`Mike Mooring <mike@divinci.ai>`).

The original commit was a modify/delete conflict — your branch modified the pre-split `format/huggingface.rs` (a single file) but main has since split it into `format/huggingface/{mod,download,publish,discovery}.rs`. The constant split landed in `huggingface/mod.rs` (`VINDEX_METADATA_FILES` / `VINDEX_BIN_FILES` / `vindex_core_files()`) and the call-site changes in `huggingface/download/mod.rs`. The progress entrypoint also went through the Model→Dataset fallback refactor from #71, so the `vindex_core_files()` iteration now sits inside the kind-probe loop. Logic is otherwise verbatim from your commit.

PR #39, opened 2026-04-26
feat(arch): DeepSeekV4Arch — V4 tensor naming (no model. prefix, ffn, w1/w2/w3)

mikeumus:
DeepSeek-V4 breaks every key convention that V3 used: no `model.` prefix, `attn` instead of `self_attn`, `ffn` instead of `mlp`, `w1`/`w2`/`w3` instead of `gate_proj`/`down_proj`/`up_proj`, `embed.weight` instead of `embed_tokens.weight`. Without this PR `larql extract` on V4-* models dies with `missing tensor: embed_tokens.weight` immediately after the safetensors header parse.

Adds a new `DeepSeekV4Arch` impl with all V4 conventions; wires through `detect.rs` so `model_type == "deepseek_v4"` dispatches to it (V2/V3 keep going to `DeepSeekArch`).

Scope: browse-tier extraction (gate + embed + down_meta). HCA / CSA forward-pass support is out of scope for this PR.

Stacks on #35 / #36 / #37 / #38.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #40, opened 2026-04-26
feat(extract): MXFP4-aware streaming gate_vectors path

mikeumus:
The streaming extract has its own `get_tensor_f32` helper (separate from the loader `tensor_to_f32` in #35). It silently returned None for any non-{F32,F16,BF16} dtype, so per-expert MXFP4 gates (I8 packed nibbles + F8_E8M0 scales) were skipped during gate_vectors.bin extraction — every layer recorded "0.0s" and the output file was 0 bytes.

Adds an I8+F8_E8M0 MXFP4 detection path at the top of `get_tensor_f32`:
- If the tensor name ends in `.weight` and dtype is I8
- And a `.scale` companion of dtype F8_E8M0 exists with matching rows
- And the cols ratio implies a sane group size {16, 32, 64, 128}
→ Unpack via `crate::format::quant::mxfp4::dequantize_expert` and return the resulting f32 `Array2` directly.

Result: `larql extract --level browse unsloth/DeepSeek-V4-Flash` now produces a populated `gate_vectors.bin` (~1 GB) instead of the previous 0 B file.

Stacks on #35 / #36 / #37 / #38 / #39. With this PR the V4 extract pipeline is end-to-end functional.


chrishayuk:
Forward-ported by #77. Authorship on the rebased commit preserved as you (`Mike Mooring <mike@divinci.ai>`).

The original commit was a modify/delete conflict — your branch modified the pre-split `extract/streaming.rs` (a single file) but main has since refactored it into `extract/streaming/{mod,tensor_io,context,stages/...}.rs`. The `get_tensor_f32` function now lives at `extract/streaming/tensor_io.rs:51`; that's where the MXFP4 detection landed.

Two other adjustments tracked in #77's body: routing the dequant call to `larql_models::quant::mxfp4::dequantize_expert` (since the fork-side `crate::format::quant::mxfp4` doesn't exist upstream), and wrapping the new `Result<Vec<f32>, ModelError>` return type into `VindexError` via `.map_err(...)?`. The detection logic itself is verbatim from your commit.

This closes the DeepSeek-V4 + MXFP4 quartet — #74 (your #35) merged, #75 (your #37) open, #76 (your #39) open, #77 (this) open.

PR #59, opened 2026-05-09
Allow pulling vindexes from HuggingFace model repos

sahuagin:
`larql publish` defaults to HuggingFace model repos, but the pull/download path only tried dataset repos. That makes public vindexes published as model repos fail to pull.

This changes the HF vindex resolver to:

- try model repos first, matching the publish default;
- fall back to dataset repos for compatibility with existing examples or older artifacts;
- use the matching cache directory prefix (`models--...` vs `datasets--...`);
- use the matching direct resolve URL for the progress/cache HEAD probe.

Observed failure before this patch:

```sh
larql pull hf://chrishayuk/gemma-3-4b-it-vindex-browse
```

failed with:

```text
Error: parse error: failed to fetch index.json from hf://chrishayuk/gemma-3-4b-it-vindex-browse
```

The repo is public as a model repo; the dataset API path is the mismatch.

Validation:

```sh
cargo fmt --check --package larql-vindex
cargo check -p larql-vindex      # passes when the FreeBSD OpenBLAS dependency fix from #58 is applied
cargo build -p larql-cli --release
./target/release/larql pull hf://chrishayuk/gemma-3-4b-it-vindex-browse
./target/release/larql show hf://chrishayuk/gemma-3-4b-it-vindex-browse
```

The pull now resolves and caches:

```text
models--chrishayuk--gemma-3-4b-it-vindex-browse/snapshots/1b2bd5adbe745891465d561c82da852d6b65529b
```

and reports the expected browse vindex metadata:

```text
34 layers, hidden_size=2560, dtype=F32, level=browse
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -154,6 +154,28 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3c07dab4369547dbe5114677b33fbbf724971019f3818172d59a97a61c774ffd"
 
 [[package]]
+name = "async-stream"
+version = "0.3.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0b5a71a6f37880a80d1d7f19efd781e4b5de42c88f0722cc13bcb6cc2cfe8476"
+dependencies = [
+ "async-stream-impl",
+ "futures-core",
+ "pin-project-lite",
+]
+
+[[package]]
+name = "async-stream-impl"
+version = "0.3.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


chrishayuk:
Forward-ported by #71 to land on top of the post round-6 download/ split + the new `RepoKind` enum on main. Authorship on the rebased commit preserved as you (`Thaddeus Covert <sahuagin@users.noreply.github.com>`). Thanks for the fix and the runtime validation against the model-repo `chrishayuk/gemma-3-4b-it-vindex-browse` — that's what made the rebase straightforward to verify.

PR #86, opened 2026-05-11
test(qwen35): Phase C.4j — chat-template prompt forward (surfaces correctness bug)

ianblenke:
## TL;DR

Even with a **properly chat-template-formatted prompt**, Qwen 3.6 27B collapses to a single high-frequency token (185983, decodes to a UTF-8 byte fragment) in this stack. The 185983 attractor now reproduces across **three distinct inputs** (token 0, raw \"Hello\", chat-template \"Hi\") — that's strong evidence of a **structural correctness bug** in C.1–C.4, not an OOD-input artifact.

## Real-run result

\`\`\`
prompt has 9 tokens: [248045, 846, 198, 12675, 248046, 198, 248045, 74455, 198]
       (<|im_start|> user \n Hi <|im_end|> \n <|im_start|> assistant \n)
prefill 0..8: forward took 2.2-2.9s per token (finite logits each step)
decode 0: argmax=185983 (9.964) top-3=[(185983, 9.96), (145263, 8.95), (225393, 8.35)]
decode 1..7: argmax=185983 every step; top-2 stabilises at 240193
assistant generated [185983]*8 → " � � � � � � � �"
\`\`\`

## Most likely culprits to investigate next

- **MRoPE handling**: Qwen 3.6 uses \`rope_dimension_sections = [11, 11, 10, 0]\`. Current code treats this as plain partial RoPE with \`rotary_dim = 32\`. For text-only inputs all 4 positions are identical so this SHOULD be equivalent — verify the actual rotation is applied to the correct dim slice.
- **Attention scale**: currently \`1/sqrt(head_dim) = 1/sqrt(256)\`. Qwen 3.6 might use \`query_pre_attn_scalar\` overrides we're missing.
- **Final-norm + lm_head**: confirm \`weights.final_norm\` actually pulls a learned \`[5120]\` vector and that \`lm_head\` shape is \`[vocab, hidden]\` post-loader.
- **GQA repeat-interleave**: K-head index for V-head h is \`h / (n_v_heads / n_k_heads)\`. Confirm this matches Qwen 3.6's conv1d output per-head ordering.

## What this PR DOES assert (same plumbing gates as C.4i)

- Tokens in vocab range
- Finite logits at every prefill + decode step
- Non-empty decode output

## Why it still lands

The test passes its plumbing gates. **Landing makes the correctness-bug signal visible and reproducible**, so the next debugging round (C.4k-ablation or C.5-correctness) starts from a known anchor point in \`main\` rather than ephemeral local state.

## Test plan

- [x] \`cargo test -p larql-inference --lib attention\` — **107/107**
- [x] \`cargo fmt --check\` clean
- [x] \`cargo clippy -p larql-inference --no-deps --all-targets\` clean
- [x] **REAL-CHAT-PROMPT**: \`LARQL_QWEN35_GGUF=Qwen3.6-27B-Q4_K_S.gguf cargo test ... real_gguf_qwen35_chat_prompt_forward --release -- --nocapture\` passes plumbing gates in 95 s


ianblenke:
Created against upstream by accident — opening on ianblenke/larql fork instead.

PR #91, opened 2026-05-13
fix(vindex): write + validate Gemma-4 PLE sidecars on --quant none (#49)

### LINKED ISSUE #49: Gemma-4 PLE tensors silently dropped on `--quant none` extracts → garbage INFER
# Gemma-4 with PLE active produces garbage INFER on `--quant none` (f16/f32) extracts

## TL;DR

`larql extract` against Gemma-4 models that activate Per-Layer Embeddings
(`hidden_size_per_layer_input > 0`, e.g. `gemma-4-E4B-it`) silently produces
a vindex with no PLE tensors when `--quant none`/`--quant f16` is used. Inference
then runs without the PLE additive contribution, producing near-random outputs
(top-5 garbage tokens at ~5-9% confidence each). The Q4_K writer already
handles this correctly — the bug is specifically in `write_f32.rs`.

## Reproducer

```bash
# Background: gemma-4-E4B-it has hidden_size_per_layer_input=256 → PLE active
larql extract google/gemma-4-E4B-it -o e4b.vindex                 # default --quant none
larql lql 'USE "e4b.vindex"; INFER "The capital of France is" TOP 5;'
```

Expected: `Paris` at high confidence.
Observed:
```
Predictions (walk FFN):
   1.  mutable             (8.92%)
   2. ッケ                   (7.39%)
   3. ceral                (7.17%)
   4.  ఉ                   (4.77%)
   5. sby                  (4.76%)
```

The same model extracted with `--quant q4k` works correctly (the Q4_K writer
already captures PLE tensors into `ple_weights.bin`).

### 1. Original safetensors contain 129 PLE tensors

```python
$ python3 -c "from safetensors import safe_open; f=safe_open('.../model.safetensors','pt'); 
              keys=list(f.keys()); ple=[k for k in keys if 'per_layer' in k.lower() or 'embed_tokens_per' in k.lower()];
              print(f'total: {len(keys)}, PLE: {len(ple)}')"
total: 2130, PLE: 129
```

The 129 are 3 globals (`embed_tokens_per_layer`, `per_layer_model_projection`,
`per_layer_projection_norm`) plus per-layer triplets (`per_layer_input_gate`,
`per_layer_projection`, `post_per_layer_input_norm`) × 42 layers.

### 2. The `--quant none` vindex's `weight_manifest.json` has zero PLE entries

```python
$ python3 -c "import json; m=json.load(open('e4b.vindex/weight_manifest.json'));
              ple=[e for e in m if 'per_layer' in e['key'].lower() or 'embed_tokens_per' in e['key'].lower() or 'layer_scalar' in e['key']];
              print(f'total: {len(m)}, PLE+layer_scalar: {len(ple)}')"
total: 505, PLE+layer_scalar: 0
```

Also note: zero `layer_scalar` entries, even though Gemma-4 forward applies
`layer_scalar` per layer when present.

### 3. `crates/larql-vindex/src/format/weights/write_q4k/mod.rs` HAS the PLE path

Lines ~474-563 explicitly:
- Write `ple_weights.bin` with `embed_tokens_per_layer.weight`,
  `per_layer_model_projection.weight`, per-layer `per_layer_input_gate.weight`,
  `per_layer_projection.weight` as `kind::TENSOR_F16`.
- Add `per_layer_projection_norm.weight`, `post_per_layer_input_norm.weight`,
  `layer_scalar` to `norms.bin`.

### 4. `crates/larql-vindex/src/format/weights/write_f32.rs` has zero such handling

A grep for `per_layer|ple_weights|embed_tokens_per|layer_scalar` returns one
match: line 589 referencing `per_layer_embed_dim` for the `model_config`
metadata only.

### 5. The forward-pass guards silently no-op on missing tensors

[`crates/larql-inference/src/forward/ple.rs:37-39`](crates/larql-inference/src/forward/ple.rs)
in `precompute_per_layer_inputs`:

```rust
let w_model_proj = match weights.tensors.get("per_layer_model_projection.weight") {
    Some(w) => w,
    None => return Vec::new(),    // silent skip
};
```

[`crates/larql-inference/src/forward/ple.rs:117`](crates/larql-inference/src/forward/ple.rs)
in `apply_per_layer_embedding`:

```rust
let per_layer_input = match per_layer_input {
    Some(p) => p,
    None => return h.clone(),    // silent skip
};
```

So when PLE tensors aren't loaded, the forward pass falls through with no
warning and no error. Garbage output is the only signal.

### 6. The CPU goldens never exercised PLE

[`crates/larql-inference/tests/test_logits_goldens.rs`](crates/larql-inference/tests/test_logits_goldens.rs)
pins `gemma4-31B-it (dense)` CPU top-1 logit at `2.366634`. But 31B's config
has `hidden_size_per_layer_input: 0` — PLE is disabled — so the codepath was
never validated by the test suite. No golden for `gemma-4-E4B-it` (PLE active)
or `gemma-4-26B-A4B-it` (MoE, also affected by a separate WalkFfn gap).

## Why this isn't CPU vs Metal

The dropped-tensor data gap will silently break Metal inference of E4B too —
Metal reads the same vindex written by the same writer. The bug is in the
extract+load pipeline, upstream of any backend selection.

### 1. `crates/larql-vindex/src/format/weights/write_f32.rs`

Add Gemma-4 sidecar writing. Mirror the existing Q4_K behavior — write 2-D
PLE tensors to `ple_weights.bin` as `kind::TENSOR_F16`, with manifest entries
for:

- `per_layer_model_projection.weight`
- `embed_tokens_per_layer.weight`
- `layers.N.per_layer_input_gate.weight`
- `layers.N.per_layer_projection.weight`

Plus add to `norms.bin`:

- `layers.N.layer_scalar`
- `layers.N.post_per_layer_input_norm.weight`
- `per_layer_projection_norm.weight`

(`post_per_layer_input_norm` belongs with norms, not with the larger PLE
tensors — same layout choice as the Q4_K writer.)

### 2. Refactor: shared helper for PLE/sidecar writing

Move the PLE-writing block from `write_q4k/mod.rs` into a shared helper in
`crates/larql-vindex/src/format/weights/` (e.g. `ple_sidecar.rs`) used by
both writers. Avoids future drift.

### 3. `crates/larql-vindex/src/format/weights/load.rs`

Teach the full float loader to load `kind::TENSOR_F16` into
`weights.tensors`, matching the Q4 loader. Then add post-load validation:

```rust
if arch.has_per_layer_embeddings() {
    require!("per_layer_model_projection.weight", weights.tensors);
    require!("embed_tokens_per_layer.weight", weights.tensors);
    for layer in 0..num_layers {
        require!(format!("layers.{layer}.per_layer_input_gate.weight"), …);
        // …per_layer_projection, post_per_layer_input_norm…
    }
    if arch.is_gemma4() {
        for layer in 0..num_layers {
            require!(format!("layers.{layer}.layer_scalar"), weights.vectors);
        }
    }
}
```

Return `VindexError::Parse` with a "rebuild this vindex with current
larql — old extract dropped Gemma-4 sidecars" message.

### 4. `crates/larql-inference/src/forward/ple.rs`

Replace the silent `return Vec::new()` / `return h.clone()` paths with
`expect()` messages. Once the loader validates the invariant, missing
tensors here is a panic-worthy bug, not a fall-through case.

The "no PLE configured" branch (line 27-29's `arch.has_per_layer_embeddings()`
check) stays — that's a real architectural state.

### 5. `crates/larql-vindex/tests/test_vindex.rs`

Synthetic non-Q4 streaming extraction test against a tiny Gemma-4 PLE config.
Asserts:

- `ple_weights.bin` exists in the output dir
- Manifest contains the expected PLE `tensor_f16` entries
- `norms.bin` contains the PLE norms + `layer_scalar` entries
- `load_model_weights` populates `weights.tensors` / `weights.vectors`

Update the existing Q4_K PLE test to also assert `layer_scalar` presence
(currently uncovered there too).

### 6. `crates/larql-inference/tests/test_logits_goldens.rs`

Add an `#[ignore]` real-vindex CPU golden for `gemma-4-E4B-it` after a fresh
extract. The current harness assumes Q4_K everywhere — branch on
`cfg.quant` so a non-Q4 vindex flows through `InferenceWeights::infer_patched`
correctly. Pin token IDs and top-1 logit (not decoded strings — the same
seed reproduces with bit-equal float behavior under the same backend).

## Adjacent issues surfaced during review

These are **separate bugs** not covered by this fix; logging here so they
don't get lost:

### A. WalkFfn doesn't implement `forward_moe_full_layer`

`crates/larql-inference/src/vindex/walk_ffn/mod.rs` has zero MoE references.
On a hybrid-MoE Gemma-4 (`gemma-4-26B-A4B-it`), the dispatch in
`predict_q4k_hidden_with_ffn`:

```rust
if weights.arch.is_hybrid_moe() {
    if let Some(h_post_attn) = run_attention_public(weights, &h, layer) {
        if let Some(h_out) = ffn_backend.forward_moe_full_layer(layer, &h_post_attn) {
            h = h_out; continue;
        }
    }
}
// fall through to dense run_layer_with_ffn
```

`WalkFfn::forward_moe_full_layer` returns `None` (default trait impl), so
MoE Gemma-4 silently falls back to the dense walk — wrong outputs.

**Fix scope:** implement `forward_moe_full_layer` on `WalkFfn` using the
existing `cpu_moe_forward` (already used by `predict_q4k_hidden`). Or, more
narrowly: route LQL INFER through `predict_q4k_hidden` directly when the
model is hybrid-MoE.

### B. `predict_with_ffn_attention` doesn't thread shared KV

Suspect `EXPLAIN INFER WITH ATTENTION` is incorrect on E4B after layers that
share KV. Out of scope here; mention in the PR for follow-up.

### C. LQL INFER tokenization doesn't chat-wrap

`exec_infer` calls `tokenizer.encode(prompt, true)` directly. The golden
harness uses `larql_inference::wrap_chat_prompt`. For -it (instruct) models,
this can produce different first-token distributions even when the model is
otherwise correct. Worth adding chat wrapping to the LQL path so a manual
`larql lql 'INFER "..."'` matches what the goldens validate.

## Why it slipped past review

- Goldens pinned only on PLE-disabled Gemma-4 (31B).
- Both the writer guard (`arch.has_per_layer_embeddings()`) and the loader
  guard (`weights.tensors.get(...)? else None`) silently no-op.
- Default extraction format on the tooling docs is implicit `--quant none`
  (the README's first example uses no `--quant` flag), so users naturally
  hit the broken path.

## Mitigation while the fix lands

`larql extract <gemma-4-with-PLE-model> --quant q4k` works today. The Q4_K
writer captures PLE; the Q4 loader hydrates it. The visualizer (or any
INFER-using consumer) can be pointed at a q4k vindex and gets correct
predictions immediately.

---

## Credits

Diagnosis by Claude Opus 4.7 (initial investigation, narrowed from "MoE bug"
hypothesis to "PLE silent drop"). Refined by Codex CLI 0.125.0 (caught the
write_f32 vs write_q4k asymmetry, plus `layer_scalar`, plus the LQL chat-wrap
hypothesis). Phase 0 mitigation (re-extract with `--quant q4k`) verified
against `google/gemma-4-E4B-it`.

deem0n:
Closes #49.

## Summary

The Q4_K writer captured Gemma-4 Per-Layer Embeddings (PLE) tensors correctly; the float writer didn't. `larql extract <model>` against `gemma-4-E4B-it` and other PLE-active variants therefore produced a vindex with zero PLE entries on the default `--quant none` / `--quant f16` path. The forward path's `precompute_per_layer_inputs` / `apply_per_layer_embedding` silently fell through (`return Vec::new()` / `return h.clone()`) when those tensors were missing — INFER returned near-random tokens at 5–9% confidence with no warning. (See #49 for the full evidence chain.)

## What changed

Fix in five parts, mirroring the Q4_K path so future quant modes inherit correctness for free.

**1. Shared `weights::ple_sidecar` module** (refactor)
Extracted `write_q4k::ple::write_ple_weights` into `weights/ple_sidecar.rs`. Both writers now call it. Same bytes on disk, same `kind::TENSOR_F16` manifest entries — pure refactor, no behavior change in the Q4_K path. The point is to prevent the bug from re-appearing: dedup means "anything fixed in one writer is automatically applied to the other."

**2. `write_f32`: call the shared helper + capture Gemma-4 norms**
After the existing per-layer norms block, the float writer now:
- Adds `arch.layer_scalar_key(layer)` to the norms loop (Gemma-4 per-layer multiplier — `None` on every other arch, so this is a true superset).
- Adds `arch.post_per_layer_input_norm_key(layer)` for PLE-active archs.
- Writes the global `per_layer_projection_norm.weight` after the final `norm.weight`.
- Calls `super::ple_sidecar::write_ple_weights` to emit `ple_weights.bin`.

The Q4_K norms writer already did all of this — the float writer dropping them was the same family of bug.

**3. `load/f32.rs`: decode `TENSOR_F16` + validate at load time**
Extended the manifest-decode match arm to handle `kind::TENSOR_F16` (mirrors `load/q4k.rs`). After the entries loop, when `arch.has_per_layer_embeddings()` is true, the loader requires every PLE tensor + vector + per-layer `layer_scalar` to be present. Missing entries fail with:

```
vindex is missing Gemma-4 PLE sidecar tensors (<N> entries, e.g. <keys>).
Rebuild with current larql — older extracts dropped these on --quant none / f16
and silently produced garbage INFER (chrishayuk/larql#49).
```

Stale (pre-fix) vindexes now fail loud with an actionable rebuild hint instead of silently producing garbage INFER.

**4. `forward/ple.rs`: replace silent fall-throughs with panics**
Replaced the three silent fall-through paths (`return Vec::new()` for the model projection, `return h.clone()` for the gate/projection tensors) with `panic!()` calls that name the missing key + reference #49. The loader enforces the invariant up front, so missing tensors here is now a load-path bug, not a fall-through case. The legitimate `!arch.has_per_layer_embeddings()` early-exit and the `per_layer_input == None` opt-out stay (those are real architectural states).

**5. Test scaffolding**
- Factored the synthetic Gemma-4 PLE fixture into a `write_gemma4_ple_fixture` helper.
- Extended it to include the per-layer `layer_scalar` 1-element vector — previously uncovered, even on the Q4_K side.
- Added `streaming_extract_noquant_carries_ple_tensors`: the direct regression test for #49. Builds the fixture, runs the real extract with `--quant none` / `ExtractLevel::Inference`, asserts `ple_weights.bin` exists, asserts the manifest has every `tensor_f16` PLE entry + every `vector` PLE norm + every `layer_scalar`, then loads via `load_model_weights` and asserts every tensor / vector key + shape.
- Extended `streaming_extract_q4k_carries_ple_tensors` with a parallel `layer_scalar` assertion.

## Out of scope

Issue #49 calls out three adjacent bugs (\"A. WalkFfn forward_moe_full_layer\", \"B. predict_with_ffn_attention shared KV\", \"C. LQL INFER chat-wrap\"). Each needs its own change set; this PR doesn't touch them.

## Test plan

- [x] `cargo test -p larql-vindex --test test_vindex -- streaming_extract_q4k_carries_ple_tensors streaming_extract_noquant_carries_ple_tensors` — 2/2 OK
- [x] `cargo test -p larql-vindex -p larql-inference --no-fail-fast` — 2342 passed, 1 failed (pre-existing macOS-local `vector_extractor_ffn_down_byte_identical` golden drift, unrelated)
- [x] `cargo test --workspace --no-default-features --no-fail-fast` — 4880 passed (+1 vs the 4879 baseline from the windows-fix run, due to the new test), same single pre-existing failure, 40 ignored
- [x] `cargo fmt -p larql-vindex -p larql-inference -- --check`
- [x] `cargo clippy -p larql-vindex --tests --no-deps -- -D warnings`
- [x] `cargo clippy -p larql-inference --lib --tests --benches --no-deps -- -D warnings`
- [x] **Cannot validate locally**: real `gemma-4-E4B-it` end-to-end extract → INFER. The synthetic regression test exercises the writer + loader + manifest paths the issue's evidence chain proved were broken; the on-disk byte layout matches the Q4_K path the issue's mitigation already confirmed correct. Someone with the model weights checked out should re-run the issue's reproducer (`larql extract google/gemma-4-E4B-it -o e4b.vindex; larql lql 'USE \"e4b.vindex\"; INFER \"The capital of France is\" TOP 5;'`) and confirm \"Paris\" appears at high confidence.


deem0n:
Seems I fixed the bug, but found another one ;-)

  ## Real-model validation on `google/gemma-4-E4B-it`

  Ran the issue's exact reproducer on the unpacked model (16 GB safetensors, PLE active per `hidden_size_per_layer_input: 256`, 42
  layers / 2560 hidden / 10240 intermediate / 262144 vocab / softcap 30.0). Extracted with this PR's build at default `--quant none`
  (`--summary-features-per-expert 64` to keep `down_meta` tractable — orthogonal to the fix). The vindex's `weight_manifest.json`
  lists every PLE sidecar entry; `ple_weights.bin` is 5.8 GB; `norms.bin` includes `layer_scalar` and the post-PLE norms.

  ```
  larql lql 'USE "/tmp/e4b.vindex"; INFER "The capital of France is" TOP 5;'
  ```

  | # | Pre-fix (per #49) | Post-fix (this PR) |
  |---|---|---|
  | 1 | `mutable` 8.92% | **`France` 99.68%** |
  | 2 | `ッケ` 7.39% | `French` 0.07% |
  | 3 | `ceral` 7.17% | `is` 0.05% |
  | 4 | `ఉ` 4.77% | `The` 0.03% |
  | 5 | `sby` 4.76% | `the` 0.02% |

  Random low-confidence garbage → coherent high-confidence English. The PLE contribution is being applied correctly across all 42
  layers (the writer-loader-forward chain is intact). Top token is `France` rather than `Paris` because the IT model fits "The capital
   of France is France" without a chat template — that's adjacent issue **C** in #49 (LQL INFER doesn't chat-wrap), out of scope here.

  ### Same prompt with explicit Gemma-4 chat wrap

  ```
  larql lql $'USE "/tmp/e4b.vindex"; INFER "<bos><|turn>user\nThe capital of France is<turn|>\n<|turn>model\n" TOP 10;'
  ```

  | # | Token | Confidence |
  |---|---|---|
  | 1 | `The` | 99.69% (starts "The capital is Paris.") |
  | 2 | **`Paris`** | 0.16% |
  | 3 | `**` | 0.15% |
  | 9 | `Berlin` | 0.0001% |

  `Paris` at rank 2, `Berlin` in top-10 — orders of magnitude above the noise floor. The PLE-driven geography knowledge is fully
  restored. The 99.69% on `The` is the IT model's preference to emit a wrapping sentence rather than the bare answer.

  ### Out-of-scope observation

  Multi-token decode via `larql run /tmp/e4b.vindex "The capital of France is" --max-tokens 5` still produces garbage
  (`ッケッケTobchal的存在`). The **first** generated token via `--max-tokens 1` is coherent — only continued decode goes wrong. That's
   a different bug (decode-time PLE bookkeeping in the KV-cached path) and doesn't share a root cause with #49. Worth a follow-up
  issue; not blocking this PR.

  ### Coverage gate

  The earlier coverage drop on `crates/larql-vindex/src/format/weights/load/f32.rs` is now back over the 75% floor — see commit
  7d50997 (`test(vindex): cover the PLE-arch missing-sidecar load rejection`) which exercises every branch of the new validation block
   end-to-end.

chrishayuk:
Hey @deem0n — main diverged from #91 enough that the original branch can't be merged cleanly (mostly the `crates/larql-inference/src/forward/ple.rs` file, which has since been collapsed into a 5-line re-export shim under ADR-0022 Step 2e2, with the PLE code moved to `larql-compute`).

I cherry-picked your two commits onto a fresh branch against current main and opened #121 as the cherry-pick PR. Both commits preserve your authorship. The only deltas from this PR:

- Dropped the changes to `crates/larql-inference/src/forward/ple.rs` (the panic-tightening) — that file is now a shim and the panic-tightening would need to land in `crates/larql-compute/src/forward/ple.rs` instead. Happy to do that as a follow-up.
- Test-file splice rebased onto main's parallel refactor of the existing Q4_K test (`KquantWriteOptions` / `load_model_weights_kquant` renames and adjacent fixture work).

Closing this in favor of #121 once that lands. Thanks for the fix — credited on both commits.

PR #94, opened 2026-05-14
feat(android): enable aarch64-linux-android cross-compilation

metavacua:
## Summary

- Gate `extern crate blas_src` behind `#[cfg(...)]` in `larql-compute`, `larql-inference`, and `larql-kv` so Android builds don't fail on a missing crate
- Move the ndarray `blas` feature from each crate's base `[dependencies]` into platform-specific `[target.cfg.dependencies]` sections — Android gets pure-Rust ndarray (matrixmultiply path), all other platforms keep their existing BLAS backends unchanged

## Build instructions (machine-local, not committed)

`.cargo/config.toml` needs NDK linker/ar entries, plus these env vars at build time:

```sh
export ANDROID_NDK_ROOT=/path/to/android-ndk-r27c
export CC_aarch64_linux_android=$NDK_BIN/aarch64-linux-android21-clang
export CXX_aarch64_linux_android=$NDK_BIN/aarch64-linux-android21-clang++
export AR_aarch64_linux_android=$NDK_BIN/llvm-ar
export OPENSSL_DIR=/path/to/openssl-android-arm64   # cross-compiled against API 21
export OPENSSL_STATIC=1
export RUSTFLAGS="-C target-feature=+dotprod"       # required by sdot in q4k kernel

cargo build --target aarch64-linux-android
```

Produces valid Android ELF64 PIE binaries (`larql`, `larql-server`, `larql-router`) confirmed with `file`.

## Notes

- The `dotprod` requirement for `q4k_q8k_dot.rs` means the binaries target ARMv8.2+ (Cortex-A55 and later, ~2017+). A proper fix would add a `#[cfg(target_feature = "dotprod")]` guard with a scalar fallback — tracked separately.
- `arm-linux-androideabi` (32-bit ARM) is blocked by Cranelift having no 32-bit ARM backend. Tracked in a separate issue.

## Test plan

- [ ] `cargo build --target aarch64-linux-android` produces ELF64 ARM aarch64 PIE binaries
- [ ] `cargo build` (host) still compiles and uses BLAS on Linux/macOS/Windows
- [ ] `cargo test` (host) passes

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


metavacua:
Pardon, my bot got confused and for some reason posted this here. I think it might be useful reference or possibly can be merged, so I'm leaving it, but if you want to close it then please do so.

PR #115, opened 2026-05-20
fix: wire --metal flag into remote FFN/MoE paths, add post-FFN norms,…

### LINKED ISSUE #114: Multi-token decode produces zeros in remote inference paths (--ffn and --moe-shards)
After extensive testing of distributed inference across two machines (Apple M1 Mac Mini + HP ProDesk i5-9500), we've identified a consistent bug affecting both the dense remote FFN path (--ffn) and the MoE remote expert path (--moe-shards): the first token from prefill is correct, but all subsequent decode tokens produce zero hidden states, causing the model to generate EOS/pad tokens repeatedly.
Environment

Client machine: Apple M1 Mac Mini, 16GB unified memory, macOS, Metal GPU
Server machine: HP ProDesk 600 G5, Intel i5-9500, 16GB DDR4, Ubuntu 26.04
Models tested: Gemma 3 4B Q4K (dense, --ffn), Gemma 4 26B A4B Q4K (MoE, --moe-shards)
larql version: ed5e9859 (latest main)

Reproduction
Dense model (--ffn):
bash# Terminal 1 - Server
larql serve gemma3-4b.vindex --ffn-only --port 8081

# Terminal 2 - Client
LARQL_RAW_PROMPT=1 larql run gemma3-4b.vindex \
  --metal \
  --ffn http://SERVER_IP:8081 \
  -n 64 \
  "The capital of France is"
MoE model (--moe-shards):
bash# Terminal 1 - Server
larql serve gemma4-26b-server.vindex --port 8081 --grpc-port 9081

# Terminal 2 - Client  
larql run gemma4-26b.vindex \
  --metal \
  --moe-shards "0-127=grpc://SERVER_IP:9081" \
  -n 64 \
  "The capital of France is"
Observed Behaviour
Both produce correct first token ("Paris") but all subsequent tokens are empty strings:
[chat] tokenised to 33 ids
[debug] tokens=64 first=["", "", ""]

  decode: 0.8 tok/s
  experts invoked: 15120  (30 layers × top-8 × 63 tokens)
Debugging Findings
Through extensive instrumentation we found:
1. moe_fn IS being called correctly — all 34 layers per decode step, server receives requests and responds.
2. Hidden state is zero after first decode step — decode_token_with_moe returns a zero vector for every decode step after the first:
[debug] decode step hidden max_abs=0.0000e0  ← every decode step
[debug] next_id=0 tok="<pad>"               ← token 0 every step
3. Prefill works correctly — all 10 prompt tokens processed with valid hidden states (max_abs ~500-300,000 range).
4. The write to new_h in moe_interleave.rs is correct — CPU raw pointer write produces valid values immediately after writing. But the value read back via read_buffer_f32(h_buf, hidden) at the end of decode_token_with_moe_fn is zero.
5. The KV cache grows correctly — kv_cache_len goes from 10 (after prefill) to 11 after first decode step.
Three Bugs Fixed (but decode still broken)
We identified and fixed three bugs in the remote FFN path:
Bug 1 — stdout not flushing (run_cmd.rs):
rust// After token print loop in run_with_remote_ffn and run_with_moe_shards:
let _ = std::io::Write::flush(&mut std::io::stdout());
Bug 2 — Missing pre-FFN norm in prefill (grid/remote_ffn.rs):
rust// Was sending raw h_post_attn to server; needs norming first:
let h_normed = apply_norm_for_ffn(weights, h_post_attn, layer);
Bug 3 — Missing post-FFN norm on server response (grid/remote_ffn.rs):
rustlet raw_out = remote.forward(layer, &x).row(0).to_vec();
apply_post_ffn_norm(weights, &raw_out, layer)
Bug 4 — --metal flag not wired into run_with_moe_shards (run_cmd.rs):
The MoE shard path always used default_backend() (CPU) regardless of --metal flag.
Bug 5 — interleaved_q4k.bin is 0 bytes for pure MoE models (pipeline_layer.rs):
build_grid_pipeline_setup panics when trying to slice into empty mmap for pure MoE models like Gemma 4 26B A4B. Fix: check if mmap is empty before slicing.
Remaining Issue
Despite fixes 1-5, the multi-token decode still produces zero hidden states. The bug appears to be in decode_token_with_moe_fn in the Metal backend — specifically how new_h is read back after handle_moe_interleave writes to it via CPU raw pointer. The value is written correctly (confirmed by reading immediately after write) but read_buffer_f32(h_buf, hidden) at line 764 of decode/mod.rs returns zeros.
Additional Findings

--metal flag not wired into run_with_remote_ffn either — same fix needed
Speed on working LAN setup: ~0.8-1.1 tok/s streaming, ~0.5 tok/s batch (gRPC)
Speed is limited by 30 sequential round trips per token, not by compute
metal-experts server feature doesn't appear to change server behaviour for --ffn-only mode

Our Fork
Our fork with bugs 1-5 fixed: https://github.com/MELDApps/larql/tree/fix/remote-ffn-output

Happy to submit a PR with the fixes once the core decode bug is resolved.

Yours,
Robert

MELDApps:
Fixes for bugs found during distributed inference testing across M1 Mac Mini + HP ProDesk.

Closes #114

## Changes
- Bug 1: Add stdout flush after token print loop in run_with_remote_ffn and run_with_moe_shards
- Bug 2: Apply pre-FFN norm before sending to remote server in prefill closure
- Bug 3: Apply post-FFN norm to server response in both prefill and decode closures  
- Bug 4: Wire --metal flag into run_with_remote_ffn (was always using CPU backend)
- Bug 4b: Wire --metal flag into run_with_moe_shards (was always using CPU backend)

Note: multi-token decode zero hidden state issue remains unresolved - see linked issue for full details.

### DIFF
--- a/crates/larql-cli/src/commands/primary/run_cmd.rs
+++ b/crates/larql-cli/src/commands/primary/run_cmd.rs
@@ -281,6 +281,7 @@ pub fn run(args: RunArgs) -> Result<(), Box<dyn std::error::Error>> {
             args.max_tokens,
             &args.ffn_dispatch,
             args.ffn_predispatch_iters,
+            args.metal,
         );
     }
 
@@ -457,7 +458,7 @@ fn run_with_moe_shards(
 
     let num_shards = configs.len();
     // Initialise compute backend early so we can report it in the topology banner.
-    let backend = larql_compute::default_backend();
+    let backend: Box<dyn larql_compute::ComputeBackend> = larql_compute::default_backend();
     eprintln!("Connecting to {} MoE shard(s)…", num_shards);
     let remote = RemoteMoeBackend::connect(configs)
         .map_err(|e| format!("failed to connect to MoE shards: {e}"))?;
@@ -528,13 +529,14 @@ fn run_with_moe_shards(
         )
     }
     .map_err(|e| format!("grid generate failed ({dispatch}): {e}"))?;
-
     for tok in &result.tokens {
         print!("{tok}");
     }
     if !result.tokens.is_empty() {
         println!();
     }
+    let _ = std::io::Write::flush(&mut std::io::stdout());
+
     let n = result.decode_ms.len();
     if n > 0 {
         let avg = result.decode_ms.iter().sum::<f64>() / n as f64;
@@ -583,6 +585,7 @@ fn run_with_remote_ffn(
     max_tokens: usize,
     dispatch: &str,
     predispatch_iters: usize,
+    metal: bool,
 ) -> Result<(), Box<dyn std::error::Error>> {
     use larql_inference::{
         generate_with_remote_ffn, generate_with_remote_ffn_batch, LayerShardedBackend,
@@ -590,7 +593,15 @@ fn run_with_remote_ffn(
     use std::time::Duration;
 
     let timeout = Duration::from_secs(ffn_timeout_secs);
-    let backend = larql_compute::default_backend();
+
+    let backend: Box<dyn larql_compute::ComputeBackend> = if metal {
+        larql_compute_metal::metal_backend()
+            .map(|m| Box::new(m) as Box<dyn larql_compute::ComputeBackend>)
+            .unwrap_or_else(|| larql_compute::cpu_backend())
+    } else {
+        larql_compute::cpu_backend()
+    };
+
     eprintln!("Connecting to remote FFN at {ffn_url}…");
     let remote = LayerShardedBackend::connect(ffn_url, timeout)
         .map_err(|e| format!("failed to connect to remote FFN server: {e}"))?;
@@ -637,13 +648,14 @@ fn run_with_remote_ffn(
         )
     }
     .map_err(|e| format!("remote-ffn generate failed ({dispatch}): {e}"))?;
-
     for tok in &result.tokens {
         print!("{tok}");
     }
     if !result.tokens.is_empty() {
         println!();
     }
+    let _ = std::io::Write::flush(&mut std::io::stdout());
+
     let n = result.decode_ms.len();
     if n > 0 {
         let avg = result.decode_ms.iter().sum::<f64>() / n as f64;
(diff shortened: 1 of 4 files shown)


chrishayuk:
Hey @MELDApps — thanks for digging into these in #114. Reviewing the diff I found three things blocking a clean merge:

1. A `break;` left mid-decode-loop in `generate_with_remote_ffn` (lines ~155-159 of the diff). That exits the loop after a single decode token — which is exactly the \"multi-token decode produces zeros\" symptom #114 documents, so I think this was bisect/debug code that didn't get cleaned up.
2. Two `eprintln!(\"[debug] ...\")` lines (also in `generate_with_remote_ffn` and `generate_with_remote_ffn_batch`).
3. The Q8K quantisation fast path (`forward_single_q8k`) was deleted from the decode `moe_fn` and replaced with the unconditional naive `forward()` call. That's a perf regression and orthogonal to the norm bug.

The other four fixes are great and I wanted to land them, so I opened #122 as a cherry-pick that keeps:

- Stdout flush in both `run_with_remote_ffn` and `run_with_moe_shards`
- `--metal` wiring in `run_with_remote_ffn`
- Pre-FFN norm in the prefill `moe_fn`
- The new `apply_post_ffn_norm` helper (with one ordering fix the unit tests caught — see PR description)

Your commit attribution is preserved on the first commit of #122. Will close this in favour of that. Happy to revisit the decode-loop norm gap (keeping the Q8K fast path) as a follow-up — that's the real fix for the multi-token zero issue.

Note for the type-ascription change in `run_with_moe_shards` — `default_backend()` is currently CPU-only (see `larql-compute/src/lib.rs:165`), so wiring `--metal` into MoE shards is more involved than the type ascription; it needs the same `if metal { metal_backend() } else { cpu_backend() }` shape we landed in `run_with_remote_ffn`. Want to send a follow-up for that?

chrishayuk:
Superseded by #122 (merged). The four legitimate fixes (stdout flush, --metal wiring, prefill pre-FFN norm, apply_post_ffn_norm helper) are in main with your attribution preserved on the first commit. Decode-loop changes (Q8K fast-path removal, debug eprintlns, mid-loop break) were dropped — they would have regressed performance and broken multi-token decode. The full decode-loop norm gap (with Q8K preserved) is still open for follow-up. Thanks again @MELDApps.

PR #118, opened 2026-05-20
test(windows): remove stale #[cfg(not(windows))] gates (follow-up to #117)

deem0n:
## Summary

Rebased on `main` per maintainer request. The CI-level fix that
landed in #124 (`OPENBLAS_NUM_THREADS=1` +
`OMP_NUM_THREADS=1` + `RUST_TEST_THREADS=1` on Windows)
eliminates the underlying OpenBLAS race on the global buffer
pool, so the two remaining stale `#[cfg(not(windows))]` test
gates that pre-date #113/#124 can come out.

## What's removed

- **`crates/larql-inference/src/trace/capture.rs`** — gate on
  `trace_final_residual_matches_raw_forward_logits` and the
  matching gate on its `forward_raw_logits` /
  `hidden_to_raw_logits` imports (which only existed to keep
  `clippy -D warnings` quiet while the test was gated).
- **`crates/larql-compute/src/backend/helpers.rs`** — gate on
  `dot_proj_gpu_some_backend_matches_fallback` and its
  "Skipped on Windows" doc paragraph.

Net diff: **2 files, 0 / −22**.

(The third gate originally targeted in this PR —
`attention_prefill_async_matches_sync` K/V check in
`larql-compute/async_compute_backend/cpu.rs` — was already
removed by #124, so it isn't in this diff.)

## Test plan

Local verification on macOS, branched off current
`chrishayuk/larql:main` (i.e. with #124's CI fix in place):

- [x] `cargo check -p larql-inference -p larql-compute --lib --tests` — clean
- [x] `cargo clippy -p larql-inference -p larql-compute --lib --tests --no-deps -- -D warnings` — clean (the previously-gated imports are now unconditionally used)
- [x] `cargo fmt -p larql-inference -p larql-compute -- --check` — clean
- [x] `trace_final_residual_matches_raw_forward_logits` — passes
- [x] `dot_proj_gpu_some_backend_matches_fallback` — passes

CI verification (the real proof — these are tests that previously
flaked or were skipped on Windows):

- [ ] `larql-inference test - windows-latest` green
- [ ] `larql-inference test - ubuntu-latest` green
- [ ] `larql-inference test - macos-14` green
- [ ] `larql-compute test · windows-latest` green
- [ ] `larql-compute test · ubuntu-latest` green
- [ ] `larql-compute test · macos-14` green
- [ ] `coverage - ubuntu` green (both larql-inference and larql-compute)

If Windows CI flakes on either of the two resurrected tests, that
would mean #124's env-var fix isn't fully covering them, and this
PR should be held — the goal is to bring Windows up to
Linux/macOS coverage, not to re-introduce flake.

### DIFF
--- a/crates/larql-compute/src/backend/helpers.rs
+++ b/crates/larql-compute/src/backend/helpers.rs
@@ -86,15 +86,6 @@ mod tests {
 
     /// `Some(CpuBackend)` → goes through trait, must equal the `None`
     /// fallback (both are CPU paths, just routed differently).
-    ///
-    /// Skipped on Windows: OpenBLAS on the Windows runner corrupts the
-    /// f32 sgemm output of the second back-to-back call (entire columns
-    /// drift by ~0.6, accompanied by `BLAS : Bad memory unallocation!`
-    /// on stderr — see PR 104 CI 2026-05-17). The trait correctness
-    /// contract is covered by the other backends in the same suite;
-    /// this test exists to pin "Some(CpuBackend) ≡ None fallback" and
-    /// there's nothing we can do at this layer about the OpenBLAS bug.
-    #[cfg(not(windows))]
     #[test]
     fn dot_proj_gpu_some_backend_matches_fallback() {
         let a = synth(4, 8, 1);
--- a/crates/larql-inference/src/trace/capture.rs
+++ b/crates/larql-inference/src/trace/capture.rs
@@ -151,11 +151,6 @@ mod tests {
     use super::*;
     use crate::ffn::FfnBackend;
     use crate::forward::trace_forward_with_ffn;
-    // `forward_raw_logits` / `hidden_to_raw_logits` are only used by
-    // `trace_final_residual_matches_raw_forward_logits`, which is gated
-    // off on Windows. Keep the imports under the same gate so `clippy
-    // -D warnings` doesn't trip on unused imports.
-    #[cfg(not(windows))]
     use crate::forward::{forward_raw_logits, hidden_to_raw_logits};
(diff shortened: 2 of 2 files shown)


chrishayuk:
Marking on hold pending #124 (the cherry-picked replacement for #117). Once #124 lands the CI gate, this can be rebased against main and the remaining stale cfg gates removed.

chrishayuk:
Hey @deem0n — looks like this branch was rebased after #124 landed but before #126 ("fix(remote-ffn): apply pre/post FFN norms across every decode-loop dispatch branch") merged a few hours ago. The current diff against main effectively reverts #126's decode-norm fixes because it removes the `prenorm_layers` / `postnorm_layers` helpers and the pre-norm call in the streaming decode `moe_fn`.

Could you rebase against current main (post-#126)? The only change you actually need here is removing the two stale `#[cfg(not(windows))]` gates from `helpers.rs` and `capture.rs`.

About the Windows test failure (`attention_prefill_async_matches_sync`): that's also worth digging into separately. The K-buffer drift signature (~0.6) is the same OpenBLAS race #124 tried to close via the env-var fix, but apparently a single-threaded BLAS + serial cargo runner isn't enough to eliminate it. May need to keep that one cfg gate after all, or find a different root cause.

chrishayuk:
Superseded by #129 (merged). The cfg-gate removals from your branch are now in main — applied against current main where #127 has already eliminated the OpenBLAS race they were workarounds for. Your branch's base had drifted enough that the diff was reverting #125 + #126 + #127, so we landed just the cfg-gate intent against current main. Thanks @deem0n.

PR #149, opened 2026-05-25
Feat/cloud clients

gburd:
Outbound LLM clients (OpenAI-compatible, AWS Bedrock) for larql-server proxy mode.


chrishayuk:
Thanks for the contribution! Closing this one for now — larql's focus is local inference on your own weights (vindex, Metal compute, quantisation). An outbound client crate for external LLM providers (OpenAI, Bedrock, etc.) feels like a different concern that doesn't fit the engine's scope, and the proxy binary duplicates the API surface already served by larql-server.

That said, I might be missing the use case you have in mind — could you share more about the vision here? If there's a concrete scenario where this integration needs to live inside larql rather than as a downstream consumer, happy to revisit.

PR #157, opened 2026-06-17
fix(server): concurrent-load OOM, startup memcheck, and inference timeout

gburd:
Three independent larql-server reliability fixes from a production incident where the first `POST /v1/infer` against an inference-level vindex on a memory-constrained host OOM-killed the process, and subsequent calls wedged the entire HTTP handler in `futex_wait`.

## 1. Concurrent-load OOM single-flight (`79254a94`)

`ensure_weights_cell` let two concurrent first-callers both run the multi-GB weight load — peak heap was 2x the weights. Now single-flighted behind an init mutex with a double-checked `OnceLock` fast path; one thread loads, the rest wait and share. Eager weight load moved into `bootstrap::serve` (before the listener binds) so a load OOM is a clean startup failure, not a first-request kill. `--lazy-weights` opt-out preserved.

## 2. Startup memory pre-flight (`5412f42e`)

New `memcheck` module + `VindexConfig::estimate_resident_bytes()`. At startup, compares the estimated resident size against the cgroup `memory.max` and refuses to bind with a one-line actionable error rather than entering a 10-second OOM-kill loop. `--no-memcheck` / `--memcheck-headroom-mib` flags.

## 3. Inference timeout + 504 (`a40f9108`)

`/v1/infer` now races the blocking forward pass against a server-side timeout (`--infer-timeout-secs`, default 60, 0 disables). On timeout it returns 504 and drops the join handle instead of leaving the handler wedged forever. `ServerError::Timeout`.

All three are BitNet-agnostic and apply to any inference-level vindex. `cargo test -p larql-server -p larql-vindex --lib`: 321 + 1081 passed.

(The lock-discipline wedge fix and the BitNet native-ternary inference path are separate follow-ups; this PR is the host-reliability subset that stands alone.)

### DIFF
--- a/crates/larql-server/examples/bench_expert_server.rs
+++ b/crates/larql-server/examples/bench_expert_server.rs
@@ -124,6 +124,7 @@ fn make_app_state(model: LoadedModel) -> Arc<AppState> {
         api_key: None,
         sessions: SessionManager::new(3600),
         describe_cache: DescribeCache::new(60),
+        infer_timeout: std::time::Duration::from_secs(60),
     })
 }
 
--- a/crates/larql-server/examples/openai_demo.rs
+++ b/crates/larql-server/examples/openai_demo.rs
@@ -123,6 +123,7 @@ fn make_app_state(model: LoadedModel) -> Arc<AppState> {
         api_key: None,
         sessions: SessionManager::new(3600),
         describe_cache: DescribeCache::new(60),
+        infer_timeout: std::time::Duration::from_secs(60),
     })
 }
 
--- a/crates/larql-server/src/bootstrap.rs
+++ b/crates/larql-server/src/bootstrap.rs
@@ -446,6 +446,23 @@ pub struct Cli {
     #[arg(long)]
     pub lazy_weights: bool,
 
+    /// Skip the startup cgroup memory pre-flight check (BUG
+    /// `infer-deadlock-oom` §5.5).  By default the server reads
+    /// `/sys/fs/cgroup/<self>/memory.max` and refuses to start when
+    /// the vindex's estimated resident size + a 512 MiB headroom
+    /// reserve exceeds the limit.  Pass `--no-memcheck` to override
+    /// (e.g. for cases where the estimate is wrong, or when running
+    /// in an environment without cgroup v2).
+    #[arg(long)]
+    pub no_memcheck: bool,
+
+    /// Headroom (MiB) to reserve below `memory.max` for the OS,
+    /// allocator overhead, and the request-handling working set.
+    /// Used by the startup pre-flight; ignored when
+    /// `--no-memcheck` is set.
+    #[arg(long, default_value_t = 512)]
+    pub memcheck_headroom_mib: u64,
+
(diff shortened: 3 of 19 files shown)


chrishayuk:
Really nice set of production fixes — and good incident write-up. Two of the three are solid as-is:

- ✅ **Single-flight load** — the double-checked `OnceLock` + init-mutex is correct: lock-free fast path, re-check under the lock, clean poison recovery (a failed load leaves the cell empty so a retry re-runs). Moving the eager load ahead of `bind` is the right call.
- ✅ **Inference timeout → 504** — semantics are honest, including the explicit note that `spawn_blocking` isn't cancellable so the thread runs to completion. `timeout=0` bypass is clean. (The sessions read-lock change is a good independent fix too.)

**One that needs a tweak before it's safe to rely on — the memcheck estimate.**
`estimate_resident_bytes()` sizes from `bytes_per_float(dtype)` (f16=2/f32=4) but ignores the `quant: QuantFormat` field — so a Q4_K/Q6_K vindex (~4.5 bits/elem) is over-counted ~3–4×, and memcheck would **refuse to start a quantized model that actually fits**. Since BitNet/quantized is the main reason to care about RSS, that undercuts the feature. Either factor `QuantFormat` into the estimate, or auto-skip memcheck (or document it as f16/f32-only) when `quant != None`. It's fine behind `--no-memcheck` for now, but worth fixing before it gates quantized deployments. (Also: `--lazy-weights` silently skips memcheck too — fine, just flag it in the help text.)

**Process nit:** these are three genuinely independent fixes (#1 and #3 are ready, #2 needs a round). If it's easy to split memcheck into its own PR, the two solid fixes can land immediately rather than waiting on the estimate work.

Heads-up: `main` is green again now — a rebase + `cargo fmt` on this PR's new files clears the CI fmt failures.

gburd:
Thanks — fixed the memcheck estimate and rebased + `cargo fmt` (clears the CI fmt failures).

**memcheck estimate** — `estimate_resident_bytes()` now returns 0 for `quant != None`, so the startup pre-flight skips quantized vindexes instead of over-counting them ~3-4× and refusing a model that fits. The bootstrap loop already treats a 0 total as skip. Dense f16/f32 — the case the estimator was validated against — still gets a real estimate. New test `estimate_skips_quantized` covers both (Q4_K → 0, dense f16 → non-zero).

I went with the skip-when-quantized option rather than modelling each format's exact resident layout — that's fragile (and for the BitNet path the weights stay packed at ~1.6 bpw, so there's no single bits/elem to plug in). Happy to switch to a per-format estimate later if you'd prefer real numbers there.

**`--lazy-weights`** — documented in the flag help that it also skips memcheck (nothing to size before a deferred load).

On the split: the estimate's now fixed so all three are ready as-is, but if you'd still rather land the two solid fixes (single-flight + timeout) separately from the memcheck work, say the word and I'll peel memcheck into its own PR.

`cargo test -p larql-server -p larql-vindex --lib`: 326 + 1087 passed.

chrishayuk:
awesome, thanks. i'll merge this in tonight, and then do any cleanups i need to post merge.  thanks again for the contribution

PR #176, opened 2026-06-29
fix(inference): clamp dense walk top-k to feature count

citizenu03bb:
Title:
fix(inference): clamp dense walk top-k to feature count

Summary:
- Clamp unlimited dense WalkFfn top-k to the layer feature count.
- Add a regression test covering `WalkFfn::new_unlimited`.

Why:
`WalkFfn::new_unlimited` previously returned `usize::MAX` from `top_k_for`
when no explicit per-layer k was configured. Dense paths that used this value
as a selection bound could request more features than exist. The unlimited
dense case should mean "all features in this layer", not an unbounded sentinel
escaping into execution.

Verification:
- cargo test -p larql-inference dense_top_k_for_clamps_to_layer_feature_count --lib
- cargo test -p larql-inference --lib

Branch:
pr/walk-ffn-dense-topk-clamp

Commit:
d75cd5d76fab898474d75fcede6ea8e04ce5578b

### DIFF
--- a/crates/larql-inference/src/vindex/walk_ffn/mod.rs
+++ b/crates/larql-inference/src/vindex/walk_ffn/mod.rs
@@ -212,7 +212,9 @@ fn walk_trace_env_enabled() -> bool {
 
 impl<'a> WalkFfn<'a> {
     fn top_k_for(&self, layer: usize) -> usize {
-        self.config.k_for(layer).unwrap_or(usize::MAX)
+        self.config
+            .k_for(layer)
+            .unwrap_or_else(|| self.index.num_features(layer))
     }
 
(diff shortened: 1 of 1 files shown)


chrishayuk:
Superseded by #206, which is this exact fix by @citizenu03bb rebased onto current `main` with the original commit authorship preserved.

This branch was 324 commits behind and unmergeable — `walk_ffn/mod.rs` had since been split, so the one-line fix stayed in `mod.rs` but the regression test moved to the extracted `dispatch_tests.rs`. No behavioural change from what you wrote.

Thanks for the patch, and sorry it sat this long.

PR #181, opened 2026-07-12
kv_dispatch: apply gemma-4 per-layer epilogue (PLE + layer_scalar) in cached prefill/decode (fixes #177, stacked on #178)

### LINKED ISSUE #177: larql run/chat KV-cache decode emits garbage on gemma-4 dense (INFER path is correct on identical input)

layer5one:
Fixes #177.

**Stacked on #178** — this branch is #178 plus one commit. The new material here is only the last commit (`1af99eb`, `crates/larql-inference/src/kv_dispatch/helpers.rs`, +71/−6). Merge #178 first and this becomes a one-file diff; alternatively merging this closes both.

Same disclosure as on #178/#179: this was debugged and verified by a Claude (AI) agent running in my homelab; I'm relaying its firsthand-verified work.

## Bug

On gemma-4 dense, single-pass `INFER` is reference-exact (after #178) but the default KV-cached generation path (`run`/`chat`) emits garbage on identical input.

Root cause: the dispatch helpers in `kv_dispatch/helpers.rs` run only attention → FFN per layer, dropping gemma-4's per-layer epilogue — `apply_per_layer_embedding` (PLE) + `apply_layer_scalar` — which gemma-4 applies on **every** layer. The non-dispatch reference path (`kv_prefill_run` / `kv_decode_step_run` in `larql-kv/src/generation.rs`) applies both, which is why `--kv-cache none` was correct while the cached engines weren't. Other architectures have neither step (both are no-ops), so existing tinymodel parity tests never caught it.

## Fix

All four helpers (sync/async × prefill/decode) now mirror the reference path:

- **Prefill** (`kv_prefill_via_dispatch`): precomputes PLE inputs from the prompt token ids and applies PLE + layer_scalar after each layer's FFN.
- **Decode**: computes PLE inputs for the single new token internally each step — no public signature changes.
- **`kv_prefill_from_hidden_via_dispatch`** (no token ids available): documented caveat — PLE is skipped, layer_scalar still applied. This entry point is only used by the Gemma-3 multimodal path today, which has no PLE, so behavior there is unchanged.

## Verification

Model: gemma-4-12B-it QAT q4_0, GGUF→vindex (artifact + scripts public at [taykso/gemma-4-12b-it-qat-q4_0-vindex](https://huggingface.co/taykso/gemma-4-12b-it-qat-q4_0-vindex)):

- Cached decode vs `--kv-cache none` on identical templated input: **token-identical** ("The capital of France is Paris.").
- Single-pass `INFER` fingerprints unchanged (chat-template prompt → `'The'` 100.00%).
- No wall-clock win at short context (the full weight pass per token dominates on this hardware); the point is correctness plus O(N) vs O(N²) scaling with context length.

## Notes

- A regression test suggestion: a dense tinymodel fixture **with** `layer_scalar` + PLE tensors, pinning `StandardEngine` output against `NoCacheEngine`. That's the gap that let this slip past the existing parity tests. Happy to add it to this PR if wanted.
- Latent sibling issue observed while root-causing, **not** fixed here: `larql-compute/src/attention/gpu.rs` (`run_attention_block_gpu`) uses the older `apply_rope_partial` and skips the per-layer effective-rope-base handling — harmless for gemma-4, wrong for rope-scaled models if that path is ever used with them. Flagging for whenever the GPU path gets attention.


chrishayuk:
Already fixed on `main` — closing as superseded rather than stale.

`21aec9be` ("fix(inference): apply PLE + layer_scalar on the standard-engine dispatch path", 2026-07-30) makes the same fix across the same four sync/async x prefill/decode loops, factored into a shared `ffn_or_moe_layer` helper rather than duplicated four ways, with bit-exact parity tests added. Your diagnosis matched it exactly: the dispatch helpers ran attention -> FFN and dropped `apply_per_layer_embedding` / `apply_layer_scalar`.

Issue #177 is not fully closed by that alone — the other half of the gemma-4 garbage is the ggml nibble layout, which is #207 (rebased from your #178). #177 should close when that lands.

Thanks for the report and the fix.

PR #196, opened 2026-07-29
feat(vindex-factory): PREFLIGHT->RELEASE build driver (§7)

chrishayuk:
## Summary
- `larql-factory::build::run_build` orchestrates FETCH → EXTRACT → SLICE → MANIFEST → VERIFY → PUBLISH → RELEASE as subprocess calls into `larql` itself, behind a `CommandRunner` trait (`SubprocessRunner` for real builds, `MockRunner` for tests) — the whole pipeline's stage ordering and failure handling is unit-tested without spawning a process or touching credentials.
- PUBLISH always uploads `--private`; RELEASE only flips a repo public once every output has verified (§8's "nothing goes public unverified"). Extends `larql-vindex`'s publish path with a `private`/visibility option (verified against the real HF OpenAPI spec) and a new `larql hf visibility` command to make that possible.
- MIRROR (R2) and REGISTER (chuk-experiments-server) stay external per the earlier scope decision — nothing in this codebase talks to either, and the spec's own text assumes both are the rig worker's job. A `BuildRecord` JSON is the hand-off point, matching `dec-bench`'s `--output-file` pattern.
- Wired in as `larql recipe build <FILE> [--scratch-dir DIR]`.
- README, ROADMAP, ROADMAP_STATUS, `docs/cli.md`, `crates/larql-factory/README.md` updated; two factual corrections landed in `docs/vindex-factory.md` itself (driver command name, VERIFY scope).

## Test plan
- [x] `cargo build -p larql-factory -p larql-cli`
- [x] `cargo test -p larql-factory --lib` (192 tests) and `cargo test -p larql-cli recipe_cmd`
- [x] `cargo clippy -p larql-factory -p larql-cli --all-targets -- -D warnings`
- [x] `cargo fmt -p larql-factory -p larql-cli -- --check`
- [x] `cargo llvm-cov -p larql-factory --lib` — every new `build/` file at or above the 90% line-coverage floor


metavacua:
You might want to put concurrency controls on the CI so that it isn't stuck running on the previous commits to the PR.

PR #205, opened 2026-08-06
fix(kv): make UnaidedTrial::observe public, clearing the larql-kv clippy gate

chrishayuk:
main is currently red. `larql-kv.yml` runs `cargo clippy -p larql-kv --all-targets --no-deps -- -D warnings`, and on `2a7c4083` that fails:

```
error: associated function `observe` is never used
  --> crates/larql-kv/src/engines/semantic_promotion/authority_regime.rs:116:19
```

`UnaidedTrial` is re-exported from `semantic_promotion`, and `record()` / `record_won()` are both `pub`, but its only constructor was `pub(crate)` — so nothing outside the type's own tests could call it, and the lib build sees it as dead.

Widening the constructor to `pub` matches the rest of the type's surface and makes the re-export actually usable, rather than silencing the lint over a type callers cannot construct.

Verified locally with the exact CI invocation: clean.

### DIFF
--- a/crates/larql-kv/src/engines/semantic_promotion/authority_regime.rs
+++ b/crates/larql-kv/src/engines/semantic_promotion/authority_regime.rs
@@ -113,7 +113,7 @@ impl UnaidedTrial {
     /// excluded. There is deliberately no path that derives it from a
     /// qualification metric: EXP-36's contrast was a decision flip, not
     /// a divergence, and a KL threshold would not have detected it.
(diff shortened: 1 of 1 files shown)


chrishayuk:
Folded into #210, which now carries this commit plus the coverage fix.

Splitting them did not work: `larql-kv.yml` runs the coverage job in the same workflow as clippy, and coverage is failing for a separate reason (#204 left five files under the 90% floor). So this PR could never show a green run on its own — its only red check was the coverage gate it does not fix. One PR, both gates, one green run.

PR #207, opened 2026-08-06
Gemma-4 dense GGUF support + ggml nibble-layout conformance (rebase of #178)

### LINKED ISSUE #177: larql run/chat KV-cache decode emits garbage on gemma-4 dense (INFER path is correct on identical input)

chrishayuk:
Rebase of #178 by @layer5one onto current `main`, plus a Q6_K NEON rework so it lands with no performance regression. Original authorship preserved across all five of the original commits; #178 was 232 commits behind and unmergeable.

## The headline finding

`main` today decodes Q4_0/Q4_1/Q5_0/Q5_1/Q6_K in a *private interleaved* nibble layout, not ggml's planar one. `quantize_q4_0` packs `block[j*2]`; `q6_k.rs` reads `ql[idx/2]`. larql therefore cannot correctly read a stock ggml Q4_0/Q6_K GGUF, and round-trip tests cannot see it because larql's own encoder writes the same wrong layout. Q4_K was already correct, which is why Q4_K_M models worked while the gemma-4 12B QAT GGUF produced structured garbage.

This PR adds ground-truth fixtures pinned to real gemma-4 GGUF bytes as decoded by gguf-py, so layout drift can no longer hide behind internal round-trip consistency.

⚠️ **MIGRATION:** vindexes containing larql-quantized Q4_0/Q6_K blocks decode incorrectly against the fixed readers and need re-extraction. #204 already forces re-extraction for `rope_scaling`, so landing this in the same window costs one re-extraction rather than two.

## Q6_K on aarch64 — resolved, no regression

The contributor's commit fixed the scalar path only, leaving the NEON and hand-asm kernels interleaved, which forced `q6k_q8k_matvec_into` through scalar for every arch — **3.47x slower** (1.739 ms vs 0.501 ms, 2560x2560) on a kernel that sits on the Q4K-direct decode path.

The final commit reworks the NEON kernel onto planar. Planar suits NEON *better* than interleaved: a 16-element scale group always lies in one nibble plane, so its `ql` and `qh` source bytes are each contiguous — the `vzip` and the four scalar byte-replicate multiplies both disappear. Measured after: **0.499 ms vs 0.501 ms for the old interleaved kernel — the full 3.47x recovered at no cost.**

`q6k_matvec_neon_matches_scalar_bit_exact` is un-ignored and passes against the planar reference; it is the gate for that kernel.

The hand-asm form is untouched and still interleaved, so the `LARQL_Q4K_ASM=1` opt-in is not offered for Q6_K and its parity test stays ignored under `TODO(q6k-planar)`. That path is opt-in and off by default.

## Rebase notes

Three of five original commits cherry-picked clean. The nibble commit needed all ten of its `larql-compute` hunks hand-ported, because `main` split `q4_common.rs` and `q4k_q8k_dot.rs` into directories. Also fixed in the port: unused-import fallout from the split, `excessive_precision` on the new gguf-py fixtures (allowed with a note — the surplus digits round to the same f32 and preserve provenance), and a `useless_vec`.

Note `q6k_matvec_asm_matches_scalar_bit_exact` **predates** #178 and was never updated by it, so #178 as authored would have gone red on Apple Silicon.

**Gates:** fmt clean; `clippy --all-targets -D warnings` clean on larql-compute, larql-models, larql-lql, larql-vindex; 944 larql-compute tests pass with 3 ignored.

Together with `main`'s `21aec9be` this should close #177.


layer5one:
Heads-up: `larql-compute-metal` on `main` has been red since this merge, and I think it is a real layout disagreement rather than a flaky test.

**Failure:** `stage_quant_matvec_routes_format_to_correct_shader`
`crates/larql-compute-metal/tests/test_kernel_vindex_integration.rs:606` — `assert!(rel < 0.02, "Q6_K route rel err {rel:.4}")`

**Likely cause:** this PR moved the CPU side of Q6_K onto ggml's planar layout (including the NEON kernel rework in f75a967d), but changed **no files under `crates/larql-compute-metal/`**. That test quantizes with `larql_compute::cpu` and reads the result back through the Metal shader, so if the Metal Q6_K path still decodes the old interleaved layout the two halves now disagree — same class as the original bug, just on the GPU side of it.

**Timeline** (`main`, `larql-compute-metal`):

| commit | time (UTC) | result |
|---|---|---|
| `2a7c4083` | 2026-08-06 23:31 | success |
| `499747f2` | 2026-08-07 00:53:24 | failure |
| `2520677f` | 2026-08-07 00:53:46 | failure |
| `aeab62b2` | 2026-08-07 00:56:43 | failure |

Green immediately before the merge, red on every run after.

**Separately, so the two do not get conflated:** the `larql-cli` clippy failure on `aeab62b2` looks unrelated to this PR. The flagged symbols (`plane_bytes`, `ParityThreshold`, `ResidualCapture`, `compare_captures`) live in `crates/larql-cli/src/commands/primary/shannon_trace/dump.rs` and `crates/larql-inference/src/residual_diff/*`, none of which this PR touched, and that job was still green at `499747f2` — it goes red starting at `2520677f`.

No patch offered for the Metal side: there is no Apple hardware here to test on, and an untested shader change is not something I would want to hand you. Happy to run anything down further on the Linux side if it helps.

Thanks for the rebase and for preserving authorship on the commits — and for tracking down `21aec9be` independently.

---
*Investigated and written by Claude (Opus 5) operating as an agent on @layer5one's infrastructure, reviewed and authorized by the account owner before posting.*

chrishayuk:
apologies.. this has been a bit of a messy one, as trying to bring in vindex3 and clean up some stuff, i'll get this fixed pretty soon

PR #211, opened 2026-08-07
fix(docs): (top level) README.md links

twistios:
### Changes
- mainly fix links in `Documentation` section
- fix other links
- also add a new link to an existing reference

### Additional notes
Other parts of the (top level) README also seem to be in need for updating, for example the current version of the Vindex(2) format spec, the existance of Vindex3 format spec, ...

I am willing to provide more pull requests for updating this and other documents. I have for example noticed that there are more documents that have links to other documents that don't work anymore, maybe because documents have been moved.

### DIFF
--- a/README.md
+++ b/README.md
@@ -520,7 +520,7 @@ LQL parser and executor. 20+ statement types across 5 categories:
 
 ## LQL Reference
 
-See [docs/specs/lql-spec.md](docs/specs/lql-spec.md) for the full language specification and [docs/lql-guide.md](docs/lql-guide.md) for a quick start guide.
+See [crates/larql-lql/docs/spec.md](crates/larql-lql/docs/spec.md) for the full language specification and [docs/lql-guide.md](docs/lql-guide.md) for a quick start guide.
 
 ### Key Statements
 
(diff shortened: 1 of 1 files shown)


chrishayuk (approved):
Verified and correct — thank you for tracking this down.

I checked every path this touches: all 8 new targets exist, and all 5 old `docs/specs/*` paths are gone from the tree. It fixes exactly the broken README links, no more and no less.

Your note about other documents was right too, and wider than you could see from README alone: **45 dead relative links across 17 tracked files.** This PR owns 7 of them; I've fixed the other 38 in a follow-up. Most were the same cause you found — docs moved from `docs/` into `crates/*/docs/` and the links stayed behind, including in `AGENTS.md`. Some pointed at things that never existed at all (the six falsified FFN backends never got per-backend pages), and those are now un-linked rather than left dangling.

The follow-up also adds `scripts/check_doc_links.py` as a blocking CI job, so this class stops recurring: it resolves every relative Markdown link in tracked files and fails on any that dangles. Running it against this branch reports *"all 693 relative documentation links resolve"* — with your change in, the tree is clean. Without it, it fails on exactly these seven, so this merges first.

Very much appreciated, and yes please to further doc PRs.

PR #222, opened 2026-08-08
feat(docs): table of contents (toc) for top level README

twistios:
A table of contents for the top level README file.

I think I got all chapters, but feel free to comment if one is missing.


twistios:
I think it would make sense to also add ToCs to other documents, for example the specifications.

PR #244, opened 2026-08-10
fix(metal): gate moe_zero_copy behind target_os = "macos"

metavacua:
## Summary

`mod moe_zero_copy;` in `crates/larql-compute-metal/src/lib.rs` (added in 3ed9623f, "perf(metal): zero-copy MoE expert dispatch") is missing the `#[cfg(target_os = "macos")]` gate every sibling module in this file has. This crate's own doc comment states it "compiles to an empty crate on non-macOS hosts — the entire implementation lives under `#[cfg(target_os = "macos")]`" — this one module breaks that invariant.

Since `larql-cli`'s `gpu` feature (which pulls in `larql-compute-metal`) is default-on, this currently breaks `cargo build -p larql-cli` on any non-macOS target with no special flags needed to hit it — plain `cargo build --release -p larql-cli` on Linux fails with:

```
error[E0432]: unresolved import `metal`
error[E0432]: unresolved import `super::buffers`
error[E0432]: unresolved import `super::moe_dispatch`
error[E0432]: unresolved import `super::MetalBackend`
error: could not compile `larql-compute-metal` (lib) due to 4 previous errors
```

(`metal`, `super::buffers`, `super::moe_dispatch` are only real under the macOS gate; `moe_zero_copy`'s own contents reference them unconditionally.)

## Fix

One line: add the missing `#[cfg(target_os = "macos")]` above `mod moe_zero_copy;`, matching every other module declaration in the file.

Confirmed no other module or crate references `moe_zero_copy` directly (`grep -rn "moe_zero_copy" --include="*.rs" .` outside `lib.rs` returns nothing), so gating it here doesn't move the break anywhere else.

## Verification

Cherry-picked onto a branch with a Linux CI build job (`cargo build --release -p larql-cli`, default features, i.e. `gpu` on) and confirmed:
- Before the fix: fails with exactly the four `unresolved import` errors above.
- After the fix: `larql-compute-metal` and `larql-cli` both compile clean, `Finished \`release\` profile [optimized + debuginfo] target(s) in 11m 32s`.

Scope is deliberately minimal — this PR touches only the missing cfg gate, nothing else.

### DIFF
--- a/crates/larql-compute-metal/examples/moss_prefill_gemm_probe.rs
+++ b/crates/larql-compute-metal/examples/moss_prefill_gemm_probe.rs
@@ -24,7 +24,7 @@ fn main() {
 
 #[cfg(target_os = "macos")]
 fn main() {
-    use larql_compute::backend::{ComputeBackend, MatMul, QuantMatVec};
+    use larql_compute::backend::{MatMul, QuantMatVec};
     use larql_compute::cpu::ops::q4_common::quantize_q4_k;
     use ndarray::Array2;
     use std::time::Instant;
--- a/crates/larql-compute-metal/src/decode/moe_interleave.rs
+++ b/crates/larql-compute-metal/src/decode/moe_interleave.rs
@@ -371,3 +371,279 @@ impl MetalBackend {
         true
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::moe_dispatch::MoeScratch;
+    use crate::MetalBackend;
+    use larql_compute::pipeline::FullPipelineLayer;
+    use larql_compute::{
+        Activation, MoeGateRule, MoeLayerWeights, MoeRoutingPolicy, MoeWeightLayout, QuantFormat,
+    };
+
+    fn backend() -> MetalBackend {
+        MetalBackend::new().expect("Metal device available on test host")
+    }
+
+    fn synth(n: usize, seed: f32) -> Vec<f32> {
+        (0..n)
+            .map(|i| (seed + i as f32 * 0.013).sin() * 0.2)
+            .collect()
+    }
+
+    fn pad_rows_to_256(data: &[f32], rows: usize, cols: usize) -> Vec<f32> {
+        let padded_cols = cols.div_ceil(256) * 256;
+        if padded_cols == cols {
+            return data.to_vec();
+        }
+        let mut out = vec![0.0f32; rows * padded_cols];
+        for r in 0..rows {
+            out[r * padded_cols..r * padded_cols + cols]
+                .copy_from_slice(&data[r * cols..(r + 1) * cols]);
+        }
+        out
+    }
+
+    /// Same layout `tests/test_kernel_moe_expert_dispatch.rs` uses for
+    /// Q4_K experts: fused `[gate | up]` halves, block-padded down rows.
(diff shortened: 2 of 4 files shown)


metavacua:
Pushed a commit adding direct unit-test coverage for the two functions failing the coverage-policy gate (`decode/moe_interleave.rs`'s `try_inline_zero_copy_moe`, and by extension `moe_zero_copy.rs`'s `encode_experts_and_combine_zero_copy`/`resolve_selected_experts`) — the merged-command-buffer "inline" fast path had zero coverage because every existing zero-copy test drives the staged/`moe_block_for_layer` entry point instead, and that path's own precondition gate (`post_expert_norm == None`, pure-MoE, identity combine) is never satisfied by `MoeRoutingPolicy::default()` or a hybrid dense+MoE layer — which is all the existing suite constructs. Also added a test for `trait_impl/mod.rs`'s untested `register_weight_region`.

Separately: while tracing why the coverage gate was even failing, found that this repo's `cargo-mutants` quality check has been silently swallowing real baseline-build failures (not just "no mutants in diff") for the exact commits that introduced this code — opened chrishayuk/larql#245 to fix that and to close the structural gap where `larql-compute-metal` is invisible to mutation testing entirely (ubuntu-only job, macOS-only crate). Not a blocker for this PR, just adjacent context that explains part of why this slipped through unnoticed for as long as it did.

metavacua:
All checks green now, including the coverage-policy gate. Final numbers, from `moe_zero_copy.rs` 57.70%/90%, `decode/moe_interleave.rs` 77.40%/90%, `trait_impl/mod.rs` 85.61%/87%, TOTAL 95.56%/96% → **`moe_zero_copy.rs` 90.33%, `decode/moe_interleave.rs` 97.94%, `trait_impl/mod.rs` 88.65%, TOTAL 96.48%**.

Four commits after the initial two got there, each fixing one thing CI's real macOS/Metal toolchain caught (no way to build or run this crate outside of it — no local reproduction was possible while writing these):
1. A Metal API misuse in my own test harness (`Command encoder released without endEncoding` — the incoming encoder must be ended/committed before calling `try_inline_zero_copy_moe`, exactly as `handle_moe_interleave` itself does).
2. A trivial `unused_assignments` clippy lint.
3. The actual bug: `moe.experts_gate_up`/`experts_down` pointed at the original un-registered byte vectors instead of slices into the registered mmap `region` — every precondition matched, but `resolve_selected_experts`'s region lookup correctly found nothing.
4. `biased_gated_servable` requires either `ClampedGlu` or empty bias arrays — a `Gated` layer with non-empty biases has no kernel, so exercising the bias-staging path needed `ClampedGlu`, not `Gated`.

Net: 4 new unit tests covering the merged-command-buffer zero-copy MoE fast path (previously untested — every existing zero-copy test only drives the staged/`moe_block_for_layer` entry point) across both grouped/non-grouped dispatch and Q4_K/Q6_K formats, plus one for `trait_impl/mod.rs`'s untested `register_weight_region`.

Also opened #245 (now green apart from an MSRV failure inherited from `main` not yet having this PR's fix — see the comment there) for a CI issue found along the way: the `cargo-mutants` quality check was silently swallowing real baseline-build failures for the exact commits that introduced this code, and is structurally unable to ever mutation-test this crate on `ubuntu-latest` regardless.

PR #245, opened 2026-08-10
ci(quality): stop masking cargo-mutants baseline failures, add a macOS mutants pass for larql-compute-metal

metavacua:
## Summary

Found while diagnosing why #244's coverage-policy gate was failing (unrelated at first glance, but the trail led straight back to this workflow):

- **The `mutants` job's baseline build silently failed** across two pushes on #241 (`feat/k3-r1-p5-moe-bias-gate-policy`, commits `3ed9623f` and `a07cf870f6`) — `moe_zero_copy.rs` wasn't cfg-gated to macOS yet (that's exactly what #244 fixes) and broke the whole-workspace build on `ubuntu-latest`. `cargo-mutants` exited **4** — ["baseline tests are already failing... no mutations were tested"](https://mutants.rs/exit-codes.html) — after finding 841 and then 1317 candidate mutants that were **never run**. Because the step is `continue-on-error: true` with its own `set +e; ...; exit 0`, both runs still showed a plain green **"cargo-mutants (informational) success"** in the PR checks tab — indistinguishable from an actual clean pass. Logs: [run 31341919447](https://github.com/chrishayuk/larql/actions/runs/31341919447/job/93317000812), [run 31342657933](https://github.com/chrishayuk/larql/actions/runs/31342657933/job/93318879078).

- **Structural blind spot, separate from the masking bug**: `larql-compute-metal` "compiles to an empty crate on non-macOS hosts" (its own doc comment) — every line inside its `#[cfg(target_os = "macos")]` modules is invisible to the ubuntu-only `mutants` job, for every PR, forever, by construction. That's exactly the crate currently failing the coverage-policy gate on #244.

## Changes

1. **`mutants`**: capture the real exit code; treat `4` (baseline failed) and `>=3` (timeout/diff-parse/internal error) as a distinct, visibly-flagged outcome (`::warning::` annotation + `$GITHUB_STEP_SUMMARY`) instead of silently `exit 0`-ing over it. Still fully non-blocking — `continue-on-error: true` is unchanged, this is still informational, just no longer misleading.
2. **New `mutants-macos`** job: same `--in-diff` pattern, scoped to `-p larql-compute-metal` on `macos-14`, so PRs touching Metal-specific code get *some* real mutation signal instead of a permanent no-op.

## Cross-repo PRs

- #244 — the PR whose coverage-gate failure led to this investigation. It now also carries direct unit tests for the specific `moe_zero_copy.rs` / `decode/moe_interleave.rs` code paths that this masked baseline could never have caught regardless (they're macOS-only, so the *old* ubuntu `mutants` job was structurally blind to them even when it wasn't outright broken).

## Test plan

- [x] YAML re-parsed with `yaml.safe_load` — valid structure, both jobs present.
- [x] Every new/changed `run:` block checked with `bash -n` — no syntax errors.
- [ ] Could not execute `cargo-mutants` itself locally (this environment has no macOS/Metal toolchain and cargo-mutants doesn't support cross-compilation) — relying on this repo's own CI to exercise both jobs for real on this PR.

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -185,33 +185,65 @@ jobs:
         run: buf build --output /dev/null
 
   mutants:
-    name: cargo-mutants (informational)
-    runs-on: ubuntu-latest
+    name: cargo-mutants${{ matrix.label }} (informational)
+    runs-on: ${{ matrix.os }}
     timeout-minutes: 45
     # PR-only: --in-diff needs a base to diff against, and a full-workspace
     # mutation run is many hours of compute.
     if: github.event_name == 'pull_request'
     continue-on-error: true
+    strategy:
+      fail-fast: false
+      # Same shape as `msrv` above, and for the same underlying reason:
+      # larql-compute-metal "compiles to an empty crate on non-macOS
+      # hosts" (its own doc comment) — every line inside its
+      # `#[cfg(target_os = "macos")]` modules is structurally invisible to
+      # a ubuntu runner, for every PR, forever. The macos-14 leg, scoped
+      # to `-p larql-compute-metal`, is the only way this workspace's
+      # mutation testing ever sees that code at all.
+      matrix:
+        include:
+          - os: ubuntu-latest
+            label: ""
+            package_args: ""
+            diff_paths: "."
+            openblas: true
+          - os: macos-14
+            label: " · larql-compute-metal"
+            package_args: "-p larql-compute-metal"
+            diff_paths: "crates/larql-compute-metal"
+            openblas: false
     steps:
       - uses: actions/checkout@v7
         with:
           fetch-depth: 0
 
+      # Computed before any install so an empty diff (e.g. a PR that
+      # never touches crates/larql-compute-metal, on the macOS leg) skips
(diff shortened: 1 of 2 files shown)


metavacua:
`MSRV · ubuntu-latest` is failing here with the same `unresolved import metal` / `unresolved import super::buffers` errors #244 fixes — not caused by anything in this PR (this PR only touches `.github/workflows/quality.yml` and adds `scripts/check_mutants_exit_status.sh`).

Root cause: `main` currently lacks #244's macOS cfg-gate fix for `moe_zero_copy.rs`. The `msrv` job's `ubuntu-latest` leg passes `--exclude larql-compute-metal`, but `larql-cli`'s `gpu` feature (default-on) pulls `larql-compute-metal` back in transitively, so excluding the crate directly doesn't actually exclude it from the build graph. Since this PR diffs against `main`, it inherits the break.

This should resolve on its own once #244 merges (or if this branch is rebased past it) — no action needed here. Confirmed everything else in this PR (including the new `cargo-mutants · larql-compute-metal` matrix leg) is green.

metavacua:
Design note: the mutants job should probably depend on the coverage jobs.

PR #260, opened 2026-08-15
VINDEX3 G6 — Metal plan lowering: Glimmer 9.4 → 18.3 tok/s with exact generation parity

chrishayuk:
Lowers a VINDEX3 `ComponentOpPlan` onto GPU-resident Metal execution. The plan
keeps the semantics; the lowerer chooses the schedule. Glimmer decode goes from
107 ms/token to 55 ms/token with **32/32 generated tokens identical**.

## Why the old path was slow

The interpreter returned to the host between every matrix operation — 209
command buffers per token. Measured cost: 215–271 µs of *empty queue* before
each dispatch, flat across a 50× range of weight bytes, collapsing to ~57 µs at
queue depth 32. Queue starvation, the same defect `test_cb_queue_starvation.rs`
convicted in the serving decoder.

Two earlier hypotheses were measured and rejected before the real one: it is not
CPU glue (7–8 ms/token) and not Metal's round-trip floor (19 µs, 4% of a
submission).

## Result

```
interpreter (209 CB/token)   107 ms/token    9.38 tok/s
lowered      (1 CB/token)     55 ms/token   18.31 tok/s

first token == p50 == p95 == 55 ms
AC replicates: 18.317 / 18.249 / 18.221 tok/s
```

The starvation signature is gone, not merely reduced. Nothing here is optimised
— no fusion, no kernel work, no layout work.

## Correctness

Gated at every rung, with each judged fact carrying a control that must dwarf
the parity residual:

| rung | parity | strongest control |
|---|---|---|
| layer fragments | 9.6e-7 | post-FFN norm omitted, 1.1M× |
| 52-layer stack | 5.4e-6 | span/position read independently |
| head | 4.6e-7 | softcap before multiplier, 184k× |
| KV evolution | 4.0e-6 | window expiry, 364k× separation |
| real container | **2.7e-7 vs interpreter** | independent f16 oracle |

The lowering's distance from the f16 oracle is *exactly* the interpreter's own
(7.382e-2 both), so the entire gap to the true model is the NVFP4 representation
and the lowering contributes nothing.

## The finding worth keeping

Anchor B (the independent oracle) caught a real omission on its first real
run: argmax 368 against the oracle's 13796, with every internal gate green. The
cause was Glimmer's weightless embedding RMS norm — a semantic operation with
**no evidencing operand**, so operand closure, four-authority G4, and every
plan-transcribed reference all agree on a model that simply lacks it.

`norm.rs` already records this class as "found by the upstream oracle, not by
any gate". It happened again, and again only the oracle saw it.

> **Operand closure cannot prove semantics that have no operand.**

## Known issue, pre-existing

`metal_decode_synthetic::attention_reaches_residual` is flaky on this branch and
on `main` (introduced in 39f39634) — a `debug_assert` in `BufferCache` firing on
allocator address reuse. Reproduced at base commit 082587c2 *without* these
changes, where it fails worse (2 failures vs 1). Not caused by this PR, but it
will make CI red intermittently and is worth its own fix.

### DIFF
--- a/crates/larql-cli/src/commands/primary/vindex3_cmd/exec.rs
+++ b/crates/larql-cli/src/commands/primary/vindex3_cmd/exec.rs
@@ -16,15 +16,28 @@
 //! seam: `--backend reference` and `--backend production` execute one
 //! program through two numerical realisations, and their dumps are
 //! directly diffable against each other as well as against upstream.
+//!
+//! Dumped runs are resumable. Each plane is written the moment its layer
+//! completes, and plane `k` is exactly the residual entering layer `k`,
+//! so the dump directory *is* the checkpoint — `--resume` reloads the
+//! last complete plane and continues bit-identically. The manifest is
+//! written only at the end and therefore doubles as the completion
+//! marker; a directory without one is an interrupted run.
 
-use std::path::Path;
+use std::path::{Path, PathBuf};
+use std::time::Instant;
 
+use larql_vindex::error::VindexError;
 use larql_vindex::format::vindex3::inspect::inspect_container;
+use larql_vindex::format::vindex3::opplan::exec::backend::PlanBackend;
 use larql_vindex::format::vindex3::opplan::exec::operands::OperandStore;
 use larql_vindex::format::vindex3::opplan::exec::production::ProductionBackend;
 use larql_vindex::format::vindex3::opplan::exec::reference::ReferenceBackend;
-use larql_vindex::format::vindex3::opplan::exec::{execute_plan, ExecutionTrace};
+use larql_vindex::format::vindex3::opplan::exec::{
+    execute_plan, execute_plan_streaming, ExecutionTrace, PlaneEvent, ResumePoint,
+};
 use larql_vindex::format::vindex3::opplan::plan_component_ops;
+use larql_vindex::format::vindex3::opplan::ComponentOpPlan;
 use ndarray::Array2;
 
 use super::super::shannon_trace::dump::{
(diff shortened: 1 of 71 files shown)


chrishayuk:
## Steps 3–5: representation frontier priced, and the 55 ms decomposed

Three more commits since the initial description.

### The frontier, under identical scheduling

Required generalising the lowering to a representation algebra
(`LoweredMatrix::{F16, Nvfp4, Mxfp4}`), so per-class policy executes under one
schedule rather than one command-buffer family per format. All arms reproduce
the interpreter's logits exactly.

```
arm               GB/tok   ppl %     pre  lowered   gain
all-NVFP4          14.91  +4.13%    9.38    18.28  1.95x
NVFP4 no-head      16.84  +3.60%    9.04    16.85  1.86x
NVFP4 FFN-only     23.21  +0.92%    8.00    13.59  1.70x
```

The endpoint spread **widened** (1.17x → 1.34x): with starvation removed the
remaining time is more purely bandwidth, so byte differences show through.
Re-fitting confirms the diagnosis from the other direction — marginal cost
barely moved, the whole win is the intercept:

```
lowered:       2.272 ms/GB + 20.8 ms fixed
pre-lowering:  2.168 ms/GB + 74.7 ms fixed
```

### MXFP4 vs NVFP4

4 interleaved repeats each on a verified-quiet machine. MXFP4 carries 5.6%
*fewer* bytes and is 3.6% **slower** — so a real kernel asymmetry survives, but
it collapsed from +21% on the starved path, meaning ~83% of the old gap was
starvation artifact. The formats cost the same to execute, so the choice is
purely quality, where NVFP4 wins outright.

### Where the 55 ms goes

Per-dispatch GPU counters are impossible here — M3 Max reports
`AtDispatchBoundary: false`. Ablating plan-optional stages instead leaves the
schedule of everything remaining intact.

```
byte term (2.272 ms/GB × 14.91 GB)   33.9 ms   61.6%
fusible glue (measured)               3.2 ms    5.8%
unexplained remainder                17.9 ms   32.6%
```

**234 dispatches of query scale + RoPE + QK norm cost nothing measurable.**
Fusing the glue buys 2–3 ms at most. The remainder is attention plus matvec
inefficiency at small shapes — attention q/o sustain 42–56 GB/s where the
756 MB head reaches 294 GB/s.

Revised near-term ceiling **~21–22 tok/s**, not 25. The levers are dispatch
geometry for the small projections, and bytes — not kernel fusion.

### Measurement note

The same binary measured **55, 84 and 90 ms** today under different contention
states. Every comparison here was run repeatedly, interleaved, on a
verified-quiet machine, because within-arm spread is the only thing that
separates signal from contention. Single runs would have produced confident
wrong answers twice.

PR #272, opened 2026-08-19
feat(models): use_double_wide_mlp / vocab_size_per_layer_input as ModelConfig fields (retire the text_features stopgap)

chrishayuk:
Follow-up to #271. The two Gemma 3n/4-E knobs move from the stopgap `inventory/text_features.rs` reader to `ModelConfig` fields read by the main parser (`detect/parser.rs`), credited via the consumed-key registry and judged by their existing carriage rules (`use_double_wide_mlp` represented only as `false`; `vocab_size_per_layer_input` a table whose width `hidden_size_per_layer_input` gates). Fields and the five literal constructors are extended at their tails so the parallel Granite `mlp_bias` change merges cleanly.

Gates: fmt/clippy clean (models, server); models 805, server 899, vindex gemma4 16 tests green; models coverage policy passes (91.05%).


chrishayuk:
Subsumed by #277 (its commit was included there and merged).

PR #282, opened 2026-08-19
fix(lql): encode mutation prompts with model architecture

citizenu03bb:
## Summary

Mutation-time prompts were tokenized directly through `tokenizer.encode(..., true)`, while query-time inference uses the model architecture-aware encoding path.

For architectures such as Gemma 4, whose tokenizer may not add BOS through its post-processor, this caused mutation residuals to be captured and calibrated in a different token coordinate system from the one used by `INFER`.

This change routes mutation prompts through the shared architecture-aware encoder.

## Changes

- Use `encode_vindex_prompt` for:
  - KNN residual capture
  - COMPOSE canonical residual capture
  - COMPOSE decoy capture
  - local post-install balancing
  - cross-fact regression checks
  - global `REBALANCE`
- Document that mutation capture and calibration must use the same prompt encoding as query-time inference.
- Add regression coverage verifying:
  - Gemma 4 receives the architecture-required BOS token when the tokenizer has no BOS post-processor.
  - legacy vindexes without a recorded model configuration retain the existing tokenizer fallback.

Target-token and entity-token encoding remain unchanged; only prompts forwarded through the model are affected.

## Why

A missing BOS token changes every downstream residual. An edit captured without BOS may therefore fail to activate, or be balanced against a prompt representation different from the one later used during inference.

Keeping prompt encoding architecture-aware and shared ensures that mutation-time and query-time residuals remain comparable.

## Compatibility

- No schema or patch-format changes.
- Existing behavior is preserved for legacy vindexes whose architecture cannot be reconstructed.
- The change applies generically through the architecture abstraction rather than branching on Gemma 4 inside the mutation code.

## Validation

- `cargo fmt --all -- --check`
- `cargo test --offline -p larql-lql --lib encode_vindex_prompt` — 2 passed
- `cargo test --offline -p larql-lql --test cov_mutation_synthetic` — 19 passed
- `cargo test --offline -p larql-lql --test cov_tier_a_synthetic insert_compose_alphabetic_tokenizer_pushes_template_decoys` — passed

## Scope

This PR is intentionally limited to mutation prompt capture and calibration. Other direct-tokenization consumers can be audited separately to keep this fix focused and reviewable.

### DIFF
--- a/crates/larql-lql/src/executor/mutation/insert/balance.rs
+++ b/crates/larql-lql/src/executor/mutation/insert/balance.rs
@@ -44,7 +44,7 @@ impl Session {
         // every magic number has a single home with its measurement-link
         // comment.
 
-        let (path, _config, _patched) = self.require_vindex()?;
+        let (path, config, _patched) = self.require_vindex()?;
         let mut cb = larql_vindex::SilentLoadCallbacks;
         let weights = larql_vindex::load_model_weights(path, &mut cb)
             .map_err(|e| LqlError::exec("balance: load weights", e))?;
@@ -52,10 +52,8 @@ impl Session {
             .map_err(|e| LqlError::exec("balance: load tokenizer", e))?;
 
         let prompt = canonical_prompt(relation, entity);
-        let enc = tokenizer
-            .encode(prompt.as_str(), true)
-            .map_err(|e| LqlError::exec("balance: tokenize", e))?;
-        let prompt_ids: Vec<u32> = enc.get_ids().to_vec();
+        let prompt_ids =
+            crate::executor::query::encode_vindex_prompt(config, &tokenizer, prompt.as_str())?;
 
         // Snapshot/restore applies only to the AMPLIFY path: when
         // UP_SCALE saturates (residual blow-up, softmax collapse in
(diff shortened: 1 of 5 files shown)


chrishayuk:
Merged to `main` in 3ab021b0 — thank you.

This was a real correctness bug, not a tidy-up: the query paths already routed through `encode_vindex_prompt`, but `balance`, `capture`, `knn` and `rebalance` called `tokenizer.encode` directly, which is exactly what that function's own doc comment forbids. On a Gemma-4 vindex that dropped BOS silently, so INSERT/REBALANCE built captures on the wrong token stream while looking fine.

Merged by hand rather than through the button because the branch had gone stale against `main`. One conflict, in `knn.rs`: `main` had since factored the prompt into `knn_canonical_prompt()`, while this PR inlined the same `format!()` and added the encoding fix. I kept **both** — the helper (its output is byte-identical) plus your `encode_vindex_prompt` routing — so nothing from either side was lost.

Verified after resolving: crate compiles, suite passes, clippy clean under the newly-pinned 1.98 toolchain, and your two new tests (`encode_vindex_prompt_prepends_gemma4_bos`, `encode_vindex_prompt_preserves_legacy_fallback`) pass by name.

PR #315, opened 2026-08-26
CPU-5/6: an integer projection path that is faster and NOT quality-validated

chrishayuk:
CPU-5 asked whether a faster CPU integer projection path preserves the model. It does not, and the rung that would have settled the surviving candidate is frozen and unrun.

## What this contains

```
df36ca9f  Q8[64] x asym-Q8[16] integer projection path      264 ms vs 348 shipped
47081a3d  bank-2 acceptance refuses an ambiguous state
fa61988a  CPU-6 paired validation protocol, banks 3A/3B, adjudicator
a87f441d  candidate / protocol / execution identities kept distinct
4a47e74a  CPU-6 parked, unrun, at frozen SHAs
```

## Read the headline carefully

`df36ca9f`'s subject says "beats production". That is a SPEED claim and the quality claim behind it did not hold:

```
quality-bank-1   PASS   KL 0.000283 vs gate 0.000316   (discovery bank)
quality-bank-2   FAIL   KL 0.000407 vs gate 0.000316   (spent)
quality-bank-3A/3B      FROZEN, NOT RUN
```

**The candidate arithmetic is quality-unvalidated.** It is reachable only through `LARQL_CPU_ARITHMETIC` and is OFF by default, so nothing in this PR changes production behaviour — but it should not be enabled anywhere on the strength of the commit subject.

Bank 2's failure is a defect in the GATE as much as in the representation: G1 was intended as "candidate degradation <= 2x shipped degradation" and implemented as "candidate KL on bank N <= 2x shipped KL measured on BANK 1" — an absolute number carried across prompt sets. Whether the candidate is relatively worse on Bank 2 is unknowable and stays unknowable; Bank 2 is spent and its missing anchor must never be measured now. CPU-6 repairs the gate and re-tests the unchanged candidate; it is ~4.6 h on a quiet machine and has never been run.

## Durable findings, independent of whether the candidate ever ships

- **Blanket Q4 x Q8 is REJECTED on Qwen3.8** — KL 0.05384, top-1 90.98%, 38 flips at BF16 margin >= 0.10.
- **Logit KL scales as the quantisation step SQUARED**, agreeing to 1%, which turns exception-set search into arithmetic. No class-level exception set rescues uniform int4: restoring 70% of the bytes removed only 45% of the damage.
- **Per-tensor int8 activation is not a usable representation.** The residual stream's peak is ~30x its RMS at depth, leaving a typical element about two bits. Blocking fixes it, 0.476 -> 0.047 rel_rms.
- **Three of five kernel hypotheses were wrong for one reason**: they removed ARITHMETIC while the binding constraint was data movement. The cheap discriminator is apparent GB/s against a known wall.

## Known debt

The crate's per-file coverage policy fails on this branch, including files this PR introduces (`integer.rs` 66%, `replay.rs` 6%, `cost.rs` 87%, `arithmetic.rs` 87%). Local coverage reads lower than CI does on ubuntu, but not by that margin. Not addressed here.

Full state: `docs/cpu-execution-roadmap.md`. Protocol: `bench/prompts/CPU6-VALIDATION.md`.


chrishayuk:
Superseded by #316, which now targets `main` directly.

`cpu5-q4q8-quality@c43306d3` is a strict ancestor of `cpu7-parallel-execution@4afbd53a`, so #316 already contains all five commits here plus three more:

- `88ab434f` docs: CPU execution roadmap
- `4a63ad70` feat(vindex3): multi-position CPU execution (CPU-7)
- `4afbd53a` fix(vindex3): zero positions is an input, not a kernel's problem

Two defects made this branch unmergeable on its own, and both are fixed above rather than here:

1. `SDOT_LANES` was `#[cfg(target_arch = "aarch64")]`-gated at `integer.rs:67` but consumed un-gated at `integer.rs:1363`, so every non-aarch64 job failed to compile `larql-vindex` with `E0425`. That cascaded into the seven failing workflows on this PR. #316 makes it `pub(super) const SDOT_LANES`.
2. CPU-7's multi-position path panicked on an empty prompt (`index out of bounds: the len is 0 but the index is 0`), breaking `generate_v3_reports_a_prefill_failure_as_a_server_error`. Fixed in `4afbd53a` with a regression test.

Repairing this branch would mean duplicating commits that already exist three ahead of it, and they would collide when #316 merges. The branch is left in place; nothing is lost.

Closing in favour of #316.

PR #422, opened 2026-09-05
feat(represent): prove sequential instance-sized codec registration (rung 2)

chrishayuk:
Rung 2 of the representation/execution contract: `BF16_ZLIB`, an entropy-coded bf16 codec that is sequential by construction, instance-sized in stored length, and registers no direct realization — the hostile witness the contract programme called for.

**Result.** Registration alone reaches the existing reference execution path: the five executor dispatch/backend files are untouched and the candidate container's production logits are bit-exact to a raw-bf16 control. A fixture produced by Python's zlib proves provider-independent lossless decoding, bf16 special values included. The one forecast contract leak, `stored_bytes(shape)`, surfaced and cost one refusal variant (`CodecError::InstanceSized`); every other contract file is byte-identical to the rung-1 merge.

**Also in this PR.** The packed-bank preflight asks the registry for row access before reading a bank, refusing sequential representations by access class with no bytes read. Seven whole-registry gates collided (six forecast); each is classified in place and generalised by declared capability rather than by label. Generalising the short-stream gate exposed a real decoder bug — a stream missing its Adler-32 trailer was accepted through the reader adapter — and whole-object decoding now requires a witnessed stream end, verified checksum, exact decoded length and no trailing bytes.

**Evidence.** Forecast frozen before code: `docs/represent/forecasts/rung2-entropy-coded-bf16.json`. Scored: `docs/represent/forecasts/rung2-execution-notes.json` (falsifier scorecard, gate classifications, findings, rung-3 handoff). Rebased onto #421; the one conflict was the no-kernel roster, where #420's direct K-quant leaves and `BF16_ZLIB` joins.

**Gates run locally after the rebase.** `cargo test -p larql-vindex` (every suite green), `cargo check --workspace --all-targets`, clippy on rustc 1.98 for aarch64 and x86_64-apple-darwin, `cargo llvm-cov` policy at 93.75% on the rebased tree (93.65% before the rebase) with every new file above the 90% floor, both documentation gates, `cargo fmt --check` last.


chrishayuk:
Merged with the `bench-regress` check red, recorded as unrelated noise under the waiver rule agreed for this PR:

- **untouched crates** — the diff changes nothing under `larql-compute-metal` or `larql-compute`, the only crates the job benchmarks;
- **historical flakiness** — the same job failed three times on the PARETO-1 PR and passed on every push to main today;
- **the run's own spread** — on the retested commit, `larql-compute-metal/quant_matvec` reports both "improved" (−56.5%) and "regressed" (+70.0%) within one run, with intervals such as [−8.9%, +69.4%] at p = 0.41;
- **every functional, coverage and platform gate green** on `ee08b35e`, including `test - windows-latest`, which the retest was for.

No threshold was weakened and no benchmark code was touched. `cargo-mutants (informational)` is not a merge gate.

PR #428, opened 2026-09-06
K3 residency vertical: the real Kimi-Linear plan through selection, budget, binding and execution (V1–V3)

chrishayuk:
The complete Kimi-Linear graph is planned without omission, selected within a multidimensional residency budget, bound and reconciled, then executed through the prepared plan with HF-oracle agreement and measured cold/warm physical behaviour.

Five commits on top of #425: the frozen forecast, V1 (semantic truth), V2 (budgeted residency), the routing correction the V3 ladder exposed, and V3 (execution and the residency curve). Forecast: `docs/represent/forecasts/k3-residency-vertical.json` (frozen, not edited); results and findings: `docs/represent/forecasts/k3-residency-vertical-notes.json`; narrative: `docs/k3-funnel.md` §4.13.

## Correctness

- **Routing fix** (`fix(exec): honor sigmoid routing, renormalisation and routed scaling`). The checkpoint declares a sigmoid router, renormalisation over the selected experts and `routed_scaling_factor` 2.446; larql-models resolved all three; the plan carried the kind and policy but had no field for the scale; the production and reference routers ran the sigmoid kind as softmax with the correction bias folded into the weights. Now: `RoutedFfnOp.branch_scale` (serialised only when present, an absent declaration executes as exactly 1), `RoutedFfnCall.branch_scale`, `production::sigmoid_select` shared by the device backend, a literal reference arm, the scale applied to every rule. Witnesses are hand-computed literals (`exec/tests/sigmoid_router.rs`) plus four fixture variants proving the scale multiplies the routed sum and leaves the shared expert alone. Kimi-Linear, DeepSeek-V3 and GLM-5.3-Flash route this way.
- **Layer parity** against the HF oracle (`scripts/kimi_generate_export.py`, per-layer taps, real container): every layer at every position cos ≥ 0.995, relL2 ≤ 0.095; position 0 relL2 ≤ 0.003 through the whole stack. Before the fix the first routed layer was at relL2 0.24–0.46.
- **Logits** at the five teacher-forced positions: KL(hf‖ours) 9.2e-5, 6.7e-5, 4.4e-4, 1.5e-3, 1.6e-3; cos ≥ 0.9999; argmax matches at every generation position, greedy trajectory `[276, 6690]` reproduced. Position 0 differs on a near-tie (margin 0.017).
- Greedy tokens had matched at 4 of 5 positions **before** the fix, with hidden states 30–45% off: a token match is not a parity witness (V3-N2).

## Architecture

- The real container's 20,159 pins select within the machine's 128 GiB and reconcile exactly against observation (declaration = observation on every operand).
- The 19,968-tensor per-expert bank binds **once** as a mapping (`MappedStored`), read through 19,968 regions of one object; 0 payload reads of the bank at binding.
- The shared expert is planned and realized, never omitted; the width is recovered from the stored gate shape where the surface predates the field.
- No new kernel: the existing mapped Kimi executor is the named realization; KDA and MLA execute through the prepared plan.

## Residency

- Seven resources kept apart: stored footprint, mapped address space, persistent resident, transient peak (max overlapping lifetime), execution touch, expected page-in per token, device memory. The budget constrains the physical working set and transient peak, never the mapping's length.
- Real container: 94.22 GB mapped versus 3.79 GB committed (observed 3.93 GB), 0.000 GB of the mapping resident after binding (mincore), 7.68 GiB per-token physical working set.
- Three decisive arms: 128 GiB selects and reconciles; 7 GiB re-selects 67 operands to a cheaper realization from the backend's own candidate lists; 3 GiB (and 5 GiB) refuses before any payload byte with the irreducible deficit and the rejected alternatives.
- Throughput (bytes touched / page-faulted per token) is admitted beside capacity. The greedy "largest resident saving" policy proves feasibility only; the multi-resource Pareto choice remains REPRESENT's and was not extended.

## Measurement

- Instrument: `larql vindex3 exec <container> --tokens … --backend production --generate N --residency-curve --repeat K` — the predicted ledger before any payload byte, then per pass and step the mapping's resident pages (mincore) and delta, minor/major fault deltas, peak RSS and time by class.
- Cold token: 0.976 GB paged in through 59,670 major faults of one 16 KiB page each (≈22 µs per fault, ≈1.1 GB/s effective — fault-bound, not bandwidth-bound) in 1.30 s; the prompt's three positions paged in 7.545 GB. The forecast's 2.94 GB/token is the disjoint bound; two thirds of the token's experts were already resident from the prompt.
- Warm token: 0 major faults; **140–430 ms observed across two runs — explicitly exploratory, not a throughput figure.** The spread is unexplained; a deterministic characterisation (same token state, per-stage timing, fault and resident-page deltas, thread scheduling, p50/p95) is the next step.
- Three residency states, not two: on disk (major fault), in the page cache but unmapped in the process (minor fault, ≈0.8 µs/page), mapped (free). The mapping's resident pages accumulate as reclaimable page cache; the budget should treat committed + transient as hard and mapped-resident as soft.

## Gates

fmt, clippy (pinned toolchain, both crates), x86_64-apple-darwin check, full larql-vindex / larql-cli / larql-models suites, per-file coverage ≥ 90% on every touched larql-vindex file (policy passes, total 93.71%), each intermediate commit builds on its own.


chrishayuk:
**Untouched-benchmark waiver (same rule as #425).**

`bench` is red on both the first run and one re-run of head `1a7d6686`. The regression check runs `larql-compute-metal: quant_matvec / matmul` and `larql-compute: linalg` against the `main` baseline on a shared macOS runner. This PR touches `larql-cli`, `larql-vindex` and docs only; its one `Cargo.lock` change adds a `windows-sys` edge that is compiled on Windows alone. Neither compute crate changes.

Measured, so the red is recorded rather than hidden:

| run | quant_matvec (24 cases) | matmul (7 cases) |
|---|---|---|
| first | +26.4%, +6.7%, +29.6% regressed beside −7.5%, −33.7% improved; the rest unchanged | — (not reached before the failure) |
| re-run | 15 regressed (+10.9% … +127.3%), 6 improved (−7.2% … −27.5%), 3 unchanged | 6 improved (−3.5% … −41.1%), 1 unchanged |

Same code moving both directions by tens of percent between two runs is runner contention, not a kernel change. Kernels and thresholds are not altered to make an architecture PR green. Merging on the functional checks.

PR #440, opened 2026-09-06
REPRESENT-PROGRESSIVE-1: one representation, three fidelities, and a pin that chooses

chrishayuk:
The first codec with more than one extent, and the machinery that lets a plan choose one. Semantic lane only — no timing, no throughput, no kernel.

Forecast frozen before implementation: `docs/represent/forecasts/represent-progressive-1.json`, with the four boundaries (available vs requested vs selected extent; semantic vs physical ownership; exact per-stream lengths; IEEE-754 edges) and the implementation order added as a dated addendum, also before code. Evidence and scored predictions: `…-notes.json`.

## What it is

`F32_PLANES` cuts an f32 image into three byte planes — `base_hi16`, `refine_8a`, `refine_8b` — at 16 / 24 / 32 bits per weight. The planes **partition** the bit pattern, so the terminal extent is the source bit for bit by construction rather than by a numerical argument. No direct kernel, deliberately: an extent must be selectable without any kernel knowing extents exist.

Four things had to become true for that to reach execution:

- **`RepresentationExtent::TERMINAL` was depth 0** — a name that read correctly while every codec had one extent and generalises backwards once a codec has three. It is `BASE` now, and a caller that means the whole representation asks the codec for `terminal_extent()`.
- **Streams stored apart had no general binding.** `streams_at(extent)` derives the streams an extent reads from the declared roles; the store binds them by declared name (`<tensor>.<stream>` siblings) and opens only those. The planner's closure pass needed the other half: `auxiliary_stream_of` gives a sibling a fate, so a container storing streams apart can be planned at all.
- **Selection had nowhere to put a depth.** `RealizationRecord.extent` carries the pin — the semantic plan stays representation-independent — with every extent the artifact offers, priced from the codec's declaration. `RepresentationFloor` states what execution requires without naming a codec or a depth, and defaults to `Exact`.
- **The ledger had no dimension an extent moves.** Under canonical decode the resident image is f32 at every depth and `touch_per_token` is priced from that image, so neither moves; the container still holds every plane, so the footprint does not move either. What a shallower extent buys is what is **opened**, and `read_to_prepare` is now that dimension, bounded by `ResidencyBudget.prepare_bytes`.

A certificate is a declaration, so `codec::conformance` checks it against the **source**: a radius the decode cannot meet, an extent that reconstructs worse than the one before it, and a deepest extent that is not the source are each refused by name. Its own refusal paths have witnesses.

## Witnesses

- **Codec-local** (14): plane order off a four-distinct-byte pattern; every extent at its declared width; the terminal extent reproducing 14 adversarial bit patterns exactly, compared as bit patterns; truncation toward zero with monotone error inside each radius; an overclaiming codec refused by depth and number; the edge semantics named one by one; a depth reading only its own planes; **one stream short and another long by the same amount refused, with the test asserting the mutated streams total exactly what the codec declares** — total-byte equality must not admit wrong per-stream extents.
- **Container-level** (5): a depth opening exactly `d+1` payloads on the store's read counter; a shallow extent decoding from a container that does not hold the deeper plane while the terminal extent refuses naming the missing tensor; and the three selection arms — generous budget pins terminal, tight budget under `Exact` **refuses** naming both deficit and requirement, tight budget under a floor with room moves the pin, with stored footprint and resident unchanged and bytes opened strictly down.
- **Out-of-tree** (2): a separate crate registers `PLANES16` — two planes, its own names, its own radius — through exported API alone; its extents reach the pin, are priced from its own declaration, move under a budget and a floor, and execute bit-exactly. No file under `src/` names it.
- **Restated universals**: two contract tests asserted what had been implemented rather than what the contract requires (the values stream by NAME; every codec having exactly one extent). Both restated, not patched.

## Scored predictions

P1 held (no trait shape change; two provided methods added). **P2 falsified** — a third structural blocker appeared, the planner's closure pass, which is the kind of thing this wave exists to find. P3 held with the generalisation the addendum anticipated (the greedy loop answers the dimension in deficit; no quality optimiser). **P4 partly falsified** — the ledger had no dimension an extent moves, which is the wave's most useful result. P5 and P6 held.

## Gates

fmt, clippy (larql-vindex and larql-cli), `x86_64-apple-darwin` check, full larql-vindex and larql-cli suites, and the per-file coverage policy (total 93.67%; `f32_planes.rs` 100%, `conformance.rs` 98.06%, `realization.rs` 95.44%, `build.rs` 93.22%, `accounting.rs` 92.90%, `operands.rs` 92.46%). **The final gate run passed.**

One earlier run of the vindex suite was red, and it is recorded rather than re-run away: `a_stage_records_its_extent_and_a_reset_clears_it` read `calls == 2`. Root cause is a **pre-existing process-global instrumentation race**, not this branch's code — `stages::LEDGER` is a `static`, and that test's mutex serialises only the three tests in its own module, so any other test in the binary opening a `Stage::Router` timer bumps the counter inside its window. It passes alone, passed on the previous head, and nothing here opens a stage timer; what changed is scheduling. It is not fixed here: the instrument belongs to the paused performance lane and produced V4's published stage split.

**Before any further qualified performance claims, that global ledger should be replaced with execution-scoped accounting and the affected measurements re-run.** Thread-local storage alone is sufficient only if a witness proves every stage scope stays on the coordinating thread.

## Not claimed

No timing, no throughput, no kernel, and no quality claim about any checkpoint: the fidelity here is a property of the encoding, certified and checked on fixtures. `F32_PLANES` is a forcing test for the contract, not a production codec — no model in this tree is stored as planes.


chrishayuk:
## Windows result, recorded

Two `test - windows-latest` jobs went red on the first pass over head `624aa59f`, and the causes were different.

**One never checked out the repository.** `actions/checkout` could not reach GitHub — `Failed to connect to github.com:443 after 21068 ms`, then `Could not resolve host: github.com`, three retries, exit 128. Nothing was compiled or run in that job.

**The other reproduced the stage-ledger race.** `a_stage_records_its_extent_and_a_reset_clears_it` failed with `calls == 11` against 1 — the same signature recorded as PR1-N5 in `docs/represent/forecasts/represent-progressive-1-notes.json`: `stages::LEDGER` is a process-global `static`, and that test's mutex serialises only the three tests in its own module, so any other test in the binary opening a `Stage::Router` timer bumps the counter inside its window. Three of the five Windows jobs passed on that same head, as did every macOS, Ubuntu and coverage job.

**Both jobs were re-run once and both passed.** The final state is every functional check green: 5/5 Windows, 5/5 macOS, 5/5 Ubuntu, 5/5 coverage, MSRV, cargo-deny (all four), cargo-audit, buf lint, doc links, no-generated-attribution. The only non-success is `cargo-mutants (informational)`, cancelled — informational, and excluded from the merge rule exactly as it was for #433.

### What is NOT concluded from the re-run

The race is not fixed and has not gone away; it was reproduced, root-caused and recorded, and this branch opens no stage timer anywhere. Isolating those three tests into their own binary would make the assertion pass while leaving the process-global instrumentation defect intact — concealing the concurrency problem rather than repairing it — so the defect stays visible and the repair stays owed.

### Prerequisite, not a suggestion

Before any further **qualified performance evidence**, the process-global stage ledger is replaced with **execution-scoped accounting** and the affected measurements are re-run. Thread-local storage alone is sufficient only if a witness proves that every stage scope remains on the coordinating thread.

PR #449, opened 2026-09-07
test(metal): witness that the lowering dispatches SiTU-GLU

chrishayuk:
## A proven kernel nobody dispatches is not an executed kernel

`tests/test_lowering_situ.rs` qualifies the `situ_glu` kernel against the scalar
authority — but it calls `bind_situ_glu` **directly**. Nothing drove
`lowering::ffn::encode_gate_up_act_from_normed`'s `FfnActivation::SituGlu` arm,
or `moe_dispatch::dispatch`'s staging arm. A lowering that ignored the plan's
combine and ran `geglu_silu` would have left every one of those tests green.

## The witness is an identity, not a second transcription

`situ_glu` computes `beta*tanh(g/beta)*sigmoid(g)*u`, with the up branch
untouched when `linear_beta` is `None`. Since `beta*tanh(g/beta) -> g`, at a
wide beta the combine **is** `silu(g)*u` — so the arm gets the same
`cpu_reference` the SiLU parity test already uses, with only
`shape.activation` changed. `run_lowered` keeps its signature and delegates to
`run_lowered_act`, so the two arms differ in nothing else.

## The pairing is what carries the claim

Stub the lowering's SiTU arm to run `geglu` and the wide-beta arm **still
passes** — it is vacuous on its own. The K3-parameter arm then reads

```
control `SiTU softcap at K3 parameters`: rel_rms 0.000e0 = 0x the parity residual
```

and fails by name. The staged-dispatch stub fails the same way
(`moved by only 0.000e0 relative against a 0.000e0 agreement floor`). Both
stubs were run and both source files restored to a zero diff.

A third arm caps only the up branch, so `has_linear` is shown to reach the
shader rather than being inferred from `linear_beta` being `Some`.

## Measured

| file | before | after |
|---|---|---|
| `lowering/ffn.rs` | 19 missed, 90.05% | **0 missed, 100.00%** |
| crate TOTAL | 959 missed, 96.04% | **941 missed, 96.12%** |

## What `moe_dispatch/dispatch.rs` cannot do, and why the witness stays

That file does not move, and no test will move it. It reports **20 missed
against 2 unique missing line numbers** — 18 are lines covered in one
instantiation and missed in another, the accounting its own policy note
describes. The 2 real ones are a defensive `break` (capability audit F10) and
the `ClampedGated` backstop, which is unreachable *by design*: it sits behind
an admission gate, and reaching it would leave the encoder unended and abort
the process.

The witness is kept because it pins the arm against a wrong-kernel edit. The
number was never the reason for it.

**This corrects a claim I made on #444.** Of the 45-line movement at
`9e2e37d9`, **19 lines were a real test gap and 26 were the instantiation
artifact** — so #446's conclusion, that the aggregate was measuring line
accounting rather than test quality, was substantially right even though it
attributed the movement to the wrong file.

## The coverage policy is deliberately untouched

`total_line_min_percent` stays where #446 put it. What floor this crate can
actually hold is a question for the **runner's own number**, not for a
projection off a local one — the runner misses 159 lines local measurement
cannot see. That is a follow-up once this PR's `test · macos-14` reports.

## Gates

15,204 workspace tests · clippy 0 · fmt clean · `larql-compute-metal` coverage
**96.12%, policy passed**.

### DIFF
--- a/crates/larql-compute-metal/src/moe_dispatch/dispatch/tests.rs
+++ b/crates/larql-compute-metal/src/moe_dispatch/dispatch/tests.rs
@@ -334,3 +334,78 @@ fn the_same_fixture_under_a_served_combine_dispatches() {
         "the fixture must dispatch"
     );
 }
+
+/// Does the STAGED dispatch reach its SiTU-GLU arm?
+///
+/// `moe_gpu_route`'s SiTU witness drives the gpu-route and zero-copy
+/// arms, and `tests/test_lowering_situ.rs` qualifies the kernel itself,
+/// but neither reaches the staging loop here — the path a layer takes
+/// when its expert bytes are supplied rather than resident in a
+/// registered region. A `match` arm here that fell through to `geglu`
+/// would leave all of those green.
+///
+/// Same shape as the dense-FFN witness: `situ_glu` computes
+/// `beta*tanh(g/beta)*sigmoid(g)*u` with the up branch untouched when
+/// `linear_beta` is `None`, and `beta*tanh(g/beta) -> g`, so at a wide
+/// beta the combine IS `silu(g)*u` and must agree with plain SiLU
+/// gating on the identical fixture. Paired with real K3 parameters,
+/// which must move the answer — the agreement alone is equally
+/// consistent with the arm never having been reached.
+#[test]
+fn situ_glu_reaches_the_staged_dispatch_with_its_parameters_bound() {
+    let Some(metal) = MetalBackend::new() else {
+        return;
+    };
+    let b = bank();
+    let s = scratch(&metal, TOP_K);
+    let x = h(23);
+    let supply = |e: usize| Some((b.gu[e].as_slice(), b.dn[e].as_slice()));
+
+    let mut silu = b.moe();
+    silu.gate_rule = MoeGateRule::Gated(Activation::Silu);
+    let plain = block(&metal, &x, &silu, &s, &supply);
+
+    // Wide beta, up branch uncapped: analytically silu(g)*u.
+    let mut wide = b.moe();
+    wide.gate_rule = MoeGateRule::SituGlu {
+        beta: 1.0e4,
+        linear_beta: None,
+    };
+    let wide_out = block(&metal, &x, &wide, &s, &supply);
+
+    assert!(wide_out.iter().all(|v| v.is_finite()));
+    let scale = plain.iter().fold(0.0f32, |m, v| m.max(v.abs())).max(1e-6);
+    let drift = plain
+        .iter()
+        .zip(&wide_out)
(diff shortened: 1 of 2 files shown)


chrishayuk:
## `bench` waiver — head `d0c0c81`, run 34098337967

Every functional check is green (`test · macos-14`, both MSRV arms, cargo-deny ×4,
cargo-audit, buf lint, doc links, both mutants arms). `bench` is the only red, and it
is the documented false-positive class.

**The diff cannot reach a benchmark.** Two files change, both test-only:

- `crates/larql-compute-metal/src/moe_dispatch/dispatch/tests.rs` — a `#[cfg(test)]`
  module, absent from the lib the bench targets link
- `crates/larql-compute-metal/tests/test_lowering_ffn_parity.rs` — an integration test

No `benches/` target, no shader, no lowering source. Both arms compile byte-identical
benchmark artifacts, which is the `noop` shape the workflow pre-registers for itself.

**Mixed sign, 25 improved vs 7 regressed.** The seven flagged cells:

| cell | change | p |
|---|---|---|
| `quant_matvec_q4_0/cpu/decode_2560` | `[+9.44% +15.82% +22.34%]` | 0.00 |
| `quant_matvec_q4_0/cpu/prefill_10240` | `[+11.04% +20.14% +30.80%]` | 0.00 |
| `quant_matvec_q4_0/metal/prefill_10240` | `[+33.82% +43.60% +52.75%]` | 0.00 |
| `quant_matvec_q4_0/cpu/lm_head_262144` | `[+3.50% +12.01% +21.38%]` | 0.01 |
| `quant_matvec_q4_k/metal/prefill_10240` | `[+7.31% +12.69% +17.94%]` | 0.00 |
| `quant_matvec_q4_kf/metal/decode_2560` | `[+23.25% +41.25% +60.59%]` | 0.00 |
| `quant_matvec_q4_kf/metal/prefill_10240` | `[+5.46% +13.42% +21.60%]` | 0.00 |

Against them, all six `q6_k` cells improved, all six `cholesky_*` improved 4–19%, all
four `ridge_decomposition_solve` cells improved up to 22%, and both `matmul_transb/M6`
pairs improved. A byte-identical binary that gets 25 cells faster and 7 slower at
p = 0.00 is measuring the runner, not the change.

**The gate's own contract is unmet.** `scripts/bench-regress.sh` never reads
`THRESHOLD=0.10`; it greps criterion's `Performance has regressed`, which fires at
criterion's 1% default. And `bench-regress.yml` states that its `noop` qualification
must pass before any red from it is treated as evidence — #446 was accidentally that
run (two lines of a JSON coverage policy, nothing compiles it) and it **failed**, with
swings of -11.4% to +15.3%.

Waived as noise on a benchmark-untouchable diff, per the established rule.

PR #482, opened 2026-09-16
perf(vindex): 3.8x faster gate KNN — reach BLAS sgemv, stop cloning f16 layers

gburd:
`describe()` on a real BitNet 2B browse container ran at ~21.6 ms/layer — about **3.3 GB/s** of effective bandwidth on a box that streams 200+. Two causes, both in the gate scan, both fixed here.

## 1. A transpose defeated BLAS

`gemv` was `matmul_transb(vec.reshape(1, hidden), gate)` — i.e. `a.dot(&b.t())` with `a` shaped `[1, hidden]`. ndarray only dispatches to BLAS when operand layouts qualify, and that shape does not reach `sgemv`.

Timed on one real layer (6912 features × 2560 dims, f32 cached):

| form | ms/layer | effective |
|---|---|---|
| `a.dot(&b.t())` (was) | 15.5 | 4.6 GB/s |
| `gate.dot(&vec)` (is) | **3.3** | **21.4 GB/s** |
| hand-rolled row dot | 9.8 | 7.2 GB/s |

**4.7× from deleting a reshape.** The diagnostic tell: a naive scalar loop beat the original "BLAS" call. If BLAS is slower than a `for` loop, it isn't being called.

## 2. f16 layers cloned the whole gate matrix per query

`gate_knn_mmap_fast` only handled f32, so every f16 layer fell through to `resolve_gate`, which ends in `cache[layer].as_ref().unwrap().clone()` — a full f32 copy of the layer. At this shape that's ~71 MB per layer per call, so a 12-layer `describe()` spent ~850 MB on allocation and memcpy *before scoring a single feature*. The decode was already cached; this was purely the copy.

Now scores out of the cache through a view. The lock is held across `gemv` — deliberate, since the alternative is cloning to release it early, which is the cost being removed. Contention is per-layer and a walk visits layers in sequence; there's a `ponytail:` note that the upgrade path is `Arc<Vec<f32>>` if high concurrency ever shows it.

## Measured end to end

Real `microsoft/bitnet-b1.58-2B-4T` browse vindex (30 layers × 6912 features), via pg_infer on PG 18.6:

| query | before | after | |
|---|---|---|---|
| `describe` (12 layers) | 288 ms | **96 ms** | 3.0× |
| `describe` (30 layers) | 648 ms | **171 ms** | 3.8× |
| `walk` | 677 ms | **198 ms** | 3.4× |
| `similar_to` | 639 ms | **198 ms** | 3.2× |
| `nearest_to` | 51 ms | **36 ms** | 1.4× |

Concurrency (`describe`, c=16): **41 → 111 qps**, p50 under load 268 → 110 ms.

The queries that *don't* scan gate vectors (`show_layers`, `show_features`, `show_relations`) did **not** move — which is the check that this fix hit what it claims rather than something else drifting.

## Correctness

Identical output: `Zone` 332.80 L22 is still the top edge for `France`, same scores to 2dp. 93 concurrent queries still return exactly one distinct answer.

`cargo test -p larql-vindex --lib`: **4736 passed, 0 failed**. `clippy --all-targets -- -D warnings` clean. `fmt` clean.

## Not addressed

The scan is now bandwidth-sane but still single-threaded (~15% of a 32-vCPU box under 8 concurrent requests) and still O(all features). `rayon` across layers looks like most of another order of magnitude. HNSW is *not* the answer at this shape — `cli.rs` says it's "break-even or net loss for dense ≤ 10K-feature models" and this one has 6912 — so a different index would be needed. Both are larger changes than this.


gburd:
## Added: rayon across layers (12× latency), and where the ceiling actually is

Second commit on this branch. `PatchedVindex::walk` scanned layers sequentially; each layer's `gate_knn` is an independent gemv, so `par_iter` over them is straightforward.

This **required** changing `f16_decode_cache` to `Mutex<Vec<Option<Arc<Vec<f32>>>>>`. The first commit's f16 fast path held the mutex across `gemv` — fine when layers are sequential, fatal once they aren't: parallel layers would serialise on one lock and cancel the change exactly. An `Arc` handle released before scoring costs a refcount bump; cloning the buffer would reinstate the ~71 MB/layer copy commit 1 removed.

A/B on a real BitNet 2B browse vindex, **alternating binaries within each rep** so drift can't masquerade as a difference, 5 reps, medians:

**Single-query latency**

| band | layers | before | after | |
|---|---|---|---|---|
| `knowledge` | 12 | 72.3 ms | **13.0 ms** | 5.56× |
| `all` | 30 | 177.1 ms | **14.4 ms** | **12.29×** |

**Throughput (qps)**

| concurrency | before | after | |
|---|---|---|---|
| 1 | 13.6 | 73.6 | 5.40× |
| 4 | 52.9 | 156.6 | 2.96× |
| 16 | 157.4 | 199.1 | 1.27× |
| 32 | 195.7 | 210.2 | **1.07×** |

The latency win is large and real. The throughput speedup **decays to 1.0** as concurrency rises, and that's the part worth recording: both binaries converge on ~200 qps. That ceiling is memory bandwidth — 200 qps × 849 MB/query ≈ 170 GB/s, at or past what this instance class sustains. Rayon fills idle cores when queries are few; at c=32 there are none to fill.

So this is a latency optimisation, and the PR text says so rather than quoting the 5.40× from c=1 as if it were a throughput number.

**Raising the ceiling needs fewer bytes per query, not more threads.** f16 scoring in place would be ~2×. Sparse/ANN retrieval over top-K features instead of a full scan is the only route to an order of magnitude — and `--hnsw` isn't it at this feature count, per `cli.rs`'s own break-even note.

## Correctness

Identical top edges on both bands (`French` 755.70 for `band=all`, `Zone` 332.80 for `knowledge`) across all 5 reps of both binaries. Verified through pg_infer on PG 18.6 too: `describe()` 13.5 ms, `walk()` 17.6 ms.

`cargo test -p larql-vindex --lib`: **4736 passed, 0 failed**. clippy `-D warnings` clean, fmt clean.

